PR #151, opened 2012-10-26
New task for Sass

pstuart2:
I love the extendability of Phing. Currently switching to it from Ant but needed a couple extra tasks. Here is one I wrote for Sass (http://sass-lang.com/).  Plan on a similar for the YUICompressor.

PR #206, opened 2013-04-18
Added Sass-Task based on PHPSass (http://www.phpsass.com)

JostBaron:


kenguest:
Is there any progress on this? Looking at #206 and #151 I would like to see these getting a fresh look - for my own purposes in work I can just <exec> against the gem installed sass tool but would prefer to see  a more mature and less error prone solution.

mrook:
I'm closing this one in favor of the work being done in #502.

PR #227, opened 2013-08-01
Test if property was set, instead of testing for non-null value

tomasfejfar:
Allows to set properties to null. Needed when from the programmatic point of view "" != null

Please test and review if it makes sense. I'm not sure if the null value test was not intentional - but for me it was a bug as it reported null-set property as undefined => it won't replace it.

### DIFF
--- a/classes/phing/parser/ProjectConfigurator.php
+++ b/classes/phing/parser/ProjectConfigurator.php
@@ -388,7 +388,7 @@ class ProjectConfigurator
     private static function replacePropertyCallback($matches)
     {
         $propertyName = $matches[1];
-        if (!isset(self::$propReplaceProperties[$propertyName])) {
+        if (!array_key_exists($propertyName, self::$propReplaceProperties)) {
             self::$propReplaceProject->log(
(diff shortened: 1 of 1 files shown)


tomasfejfar:
:see_no_evil: I just found out that I completly forgot to bump this ✊ ... three years ago :trollface: But I think it [still makes sense to fix it](https://github.com/phingofficial/phing/blob/master/classes/phing/parser/ProjectConfigurator.php#L391). Should I rebase on top of current master?

mrook:
Sorry I never bothered to review or comment the PR :(

I think it's a good change, but I'm thinking if this constitutes a BC break. What do you think?

tomasfejfar:
Yes, I'm afraid someone might have used setting null in place of unsetting the property. 

When do you plan next major release? It could also be switchable, maybe - but there is no precedent for it, is there?

mrook:
The next major release will be 3.0 and should be out either late this year or early next year. Could you rebase this PR against the 3.0 branch?

tomasfejfar:
Great 👍

PR #228, opened 2013-08-05
Enable HttpRequesTask to make POST requests

tsphethean:
This PR allows the HTTP method to be set as POST (rather than the default GET) and for post parameters to be added as per the documentation of HTTP_Request2 at http://pear.php.net/manual/en/package.http.http-request2.request.php

An example of how an http-request task could now look: 

``` xml
<http-request
      url=""
      authUser=""
      authPassword=""
      verbose="true"
      method="POST"
      >
      <config name="adapter" value="HTTP_Request2_Adapter_Curl"/>
      <header name="user-agent" value="Phing HttpRequestTask"/>
      <header name="content-type" value="application/json" />
      <postparameter name="param1" value="test1" />
      <postparameter name="param2" value="test2" />
    </http-request>
```

Future patches could add support for other HTTP methods.

### DIFF
--- a/docs/docbook5/en/source/appendixes/optionaltasks.xml
+++ b/docs/docbook5/en/source/appendixes/optionaltasks.xml
@@ -3201,6 +3201,13 @@ Note that you can omit both startpoint and track attributes in this case
                             disconnect</entry>
                         <entry>No</entry>
                     </row>
+                    <row>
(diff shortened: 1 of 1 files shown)


mreiche:
Does this support posting application/json in POST request body? The official documentation is not specific enough.

PR #243, opened 2013-10-10
Fix for ticket #309 (meaning/definition of project.basedir)

mpdude:
See http://www.phing.info/trac/ticket/309

This might be considered a BC break. Acutally it reverts behaviour to what it was many moons ago.
The fix was (IMO) wrong in the first place, improved docs (including docbook) instead.

(cherry-picking eaf81bc316ed965f4d57b66c75a65e5dbf9e0c8c off unstable-3.0)


mrook:
@mpdude Yes, I think I still want this in 3.0 - but it'll need to be rebased against the new 3.0 branch.

mpdude:
Bummer, I just screwed this PR because I pushed the variant suitable for the 3.0 branch into it, but the base is wrong. Will open a new PR.

PR #261, opened 2014-01-25
Consider the strings "1" and "0" to be true and false, respectively.

HuffAndPuff:
Somewhere along a build process PHP has written a boolean `true` to a property file as the string `"1"`. It would be of great benefit to me if the strings "1" and "0" could be recognized as boolean values by StringHelper.

This is just a suggestion, I haven't investigated it's potential effects but it seems harmless to me.

### DIFF
--- a/classes/phing/util/StringHelper.php
+++ b/classes/phing/util/StringHelper.php
@@ -12,8 +12,8 @@
  */
 class StringHelper {
 
-    private static $TRUE_VALUES = array("on", "true", "t", "yes");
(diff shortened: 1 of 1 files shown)


kenguest:
Out of curiosity, I did a quick check, running xmllint against a small example of "<delete verbose='1' ...>" and validating with the phing-grammar.rng file, and found no errors were raised. So there's nothing to worry about in that regard :+1:

mrook:
Closing in favor of #590.

PR #292, opened 2014-07-14
Added extra options

damiandennis:
Added extra options to support a single path, also added option to use an excludePath to filter the path supplied, so that this can be updated in the repo instead of manually on the server.


mrook:
@damiandennis can you clean up the PR and make the test(s) pass?

mrook:
Closing for lack of feedback.

PR #310, opened 2014-10-04
PharDataTask

siad007:
Alternative archive maker task without a dependency like ArchiveTar but a minimum PHP version of 5.3 or the phar pecl is required.

http://php.net/manual/en/class.phardata.php


mrook:
Unit test(s)? ;-) Also needs a rebase.

siad007:
Added unit tests and rebased.

PR #313, opened 2014-10-06
Added PropertyRegexTask

siad007:
Performs regular expression operations on an subject string, and sets the results to a property.
Inspired by http://ant-contrib.sourceforge.net/tasks/tasks/propertyregex.html


mrook:
Thanks, would you mind adding unit test(s) as well?

siad007:
@mrook tests were added.

PR #351, opened 2014-12-30
Added IsPropertyFalse/-True conditions

siad007:
Added `<ispropertytrue property="foo"/>` and `<ispropertyfase property="foo"/>` condition.

```
        <property name="test" value="yes" />
        <if>
            <ispropertytrue property="test"/>
            <then>
                <echo message="foo"/>
            </then>
        </if>
        <if>
            <ispropertyfalse property="test"/>
            <then>
                <echo message="bar"/>
            </then>
        </if>
```

### DIFF
--- a/classes/phing/tasks/system/condition/ConditionBase.php
+++ b/classes/phing/tasks/system/condition/ConditionBase.php
@@ -164,6 +164,28 @@ abstract class ConditionBase extends ProjectComponent
     }
 
     /**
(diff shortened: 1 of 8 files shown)


mrook:
Can you add unit test(s)?

siad007:
@mrook tests added

mpdude:
See http://www.phing.info/trac/ticket/495 which is related

PR #353, opened 2015-01-07
Added support for status code on exit and updated FailTask.

siad007:
Extends phing with status code on shutdown and updated FailTask to support status flag and nested condition.
Example:
```
<fail status="255">
    <condition>
        <not>
            <isset property="thisdoesnotexist"/>
        </not>
    </condition>
</fail>
```


siad007:
There you are :smile:

PR #358, opened 2015-01-13
Make basedir property (including its default value) a path relative to the buildfile

mpdude:
See http://www.phing.info/trac/ticket/309

This might be considered a BC break. Acutally it reverts behaviour to what it was many moons ago.
The fix was (IMO) wrong in the first place, improved docs (including docbook) instead.

(cherry-picking eaf81bc316ed965f4d57b66c75a65e5dbf9e0c8c off unstable-3.0)
(This also obsoletes #243)


mrook:
Cherry-picked dfdb0bc8095db18284de364b421d320be3c1b6fb to master.

PR #363, opened 2015-01-17
Deferred & refactored property expansion. Property file sections. Support property arrays.

mpdude:
This picks 1a0e8674ee39b10ffd2c81c123d79130e61fa8dd, 22ad200e5276f8d2f3c0ec629e142c57ade3e107 and a05f680fe3c0b0b9c21409850428eee0ad08e452 from the `unstable-3.0` branch and applies it to 3.0. Unfortunately this was a series of commits mixing up different features :-(.

The merge was performed almost completely manually so hopefully nothing is lost.

Also see http://www.phing.info/trac/ticket/320 for some background information.

**Deferred property expansion** allows you to do really useful things like
```
# Base property file
webdir = web
cachedirname = cache
cachedir = ${webdir}/${cachedirname}

# Load another property file with
webdir = www
# ... or even pass -Dwebdir=docroot on the command line
```
This will finally expand to `docroot/cache`. So it gives you the power to come up with several layers of configuration/property files where later levels can still apply changes to (composite) patterns defined earlier.

**Property file sections** work like the following
```
some.global = value
[development]
settings.show_errors = true
[staging]
settings.show_errors = false
[production : staging]
caching.enabled = true
```
I will update the docs later on, but this should give you the basic idea. A section can be chosen when loading the properties file. Sections can inherit from other ones, possibly overriding values.

**Property arrays** allow you to load different properties files like
```
assets.include[] = some/path
assets.include[] = some/other/path
```
This will be merged into a single `assets.include` property which is an array and can be used to work a set of user-supplied paths.
 
Backwards Compatiblity?
---
* Does not change public method signatures (unless I missed something)


mpdude:
@mrook I think I am done with this for now. There are still some additions/extensions for this coming later on the `unstable-3.0` branch, but I think we could merge these features independently if you like them.

Feel free to ask me your questions and I'll see if I can remember why I did things that way 3.5 years ago :).

mpdude:
Oh, the Travis build failed and I cannot see why...?

mpdude:
@mrook How can we get along with this?

mrook:
@mpdude rebase and apply? I think you got most of it.

mpdude:
Sounds good. There are still some additional refactorings and goodies I'd like to pick from `unstable-3.0`, but this PR is already huge enough so I'll open new ones for the other stuff.

mpdude:
Rebased. I don't understand what's wrong with Travis, can you help @mrook?

mrook:
Might be a missing class / use statement somewhere. I've noticed earlier that the error reporting seems to be somewhat broken. I'll have a look!

mrook:
@mpdude your code is throwing RuntimeException's which are swallowed by bin\phing.php, I've added a fix for that here: a15762153d35780a34a8934771377f272f536480

At some point we may want to improve that exception handling code.

mpdude:
Hooray, tests passing again!

PR #366, opened 2015-01-21
Added isFailure condition

siad007:
Example:
```
    <target name="test">
		<exec command="test" returnProperty="return.code"/>
		<if>
			<isfailure code="${return.code}"/>
			<then><echo msg="${return.code}"/></then>
		</if>
	</target>
```


siad007:
Tests added.

PR #369, opened 2015-01-23
Use an absolute path (using __DIR__) in test/bootstrap.php

mpdude:
This makes it possible [for me, at least] to run PHPUnit from my IDE.

### DIFF
--- a/test/bootstrap.php
+++ b/test/bootstrap.php
@@ -1,6 +1,6 @@
 <?php
 
-require_once __DIR__ . '/../vendor/autoload.php';
+require_once dirname(__FILE__) . '/../vendor/autoload.php';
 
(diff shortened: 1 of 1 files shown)


siad007:
@mpdude Does it not make more sense to use `dirname(__FILE__)` instead? `__DIR__` is >= PHP 5.3 .

mpdude:
To provide PHP 5.2 support?

siad007:
yes :) or did i miss something...

mpdude:
No, probably not... It's just that I though I was the last man on earth still running 5.3 so...

mrook:
This is a PR against the 3.0 branch which is 5.4+, __DIR__ can be used.

PR #371, opened 2015-01-24
[WIP] Added DefaultExcludes Task

siad007:
Added ant like DefaultExcludes task as described on https://ant.apache.org/manual/Tasks/defaultexcludes.html

### DIFF
--- a/classes/phing/Project.php
+++ b/classes/phing/Project.php
@@ -107,13 +107,33 @@ class Project
     /**  Build listeneers */
     private $listeners = array();
 
+    /** @var DefaultExcludesContainer $globalExcludes */
+    private $globalExcludes;
+
     /**
      *  Constructor, sets any default vars.
      */
     public function __construct()
     {
-        $this->fileUtils = new FileUtils();
-        $this->inputHandler = new DefaultInputHandler();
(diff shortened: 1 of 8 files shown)


mpdude:
The way this has to modifiy the static properties of `DirectoryScanner` is somewhat ugly, difficult to test and has some serious "WTF?" potential.

A quick scan shows that `DirectoryScanner` instances are created in `AbstractFileSet`, `Excludes` and `PearPackageFileSet`. In all these locations, a `Project` instance is nearby.

So what do you think about
* making the default excludes a property of the `Project` and passing them to the scanner instance when needed
* or make `Project` directly hold a `DirectoryScanner` instance that can be configured by this task?

Downside of the first one is that it makes the `Project` some data it does not really need to be aware of, just to pass it on later. 

The second one only works if the `DirectoryScanner` is stateless (haven't checked) or the `Project` rather works as a factory and returns new instances. Also we'd need to see how `PearPackageFileSet` fits into the picture.

What do you think?

mpdude:
Hint: PearPackageScanner only uses `set(In|Ex)cludes` and `is(In|Ex)cluded` from `DirectoryScanner`. We surely can refactor that, no need for inheritance.

PR #372, opened 2015-01-24
Add first tests for Phing\Project

mpdude:
Well, really some first tests only.

### DIFF
--- a/test/ProjectTest.php
+++ b/test/ProjectTest.php
@@ -44,7 +44,6 @@ class ProjectTest extends \PHPUnit_Framework_TestCase
 
     public function testGetPropertiesWillExpandPropertyValues()
     {
-        $this->markTestSkipped('This is currently not supported in Phing but might change with https://github.com/phingofficial/phing/pull/363');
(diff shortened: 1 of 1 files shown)


test/ProjectTest.php
```diff
@@ -0,0 +1,50 @@
...
+
+        $this->assertEquals(array('foo' => 'foo', 'bar' => 'bar'), $this->project->getProperties());
+    }
+
+    public function testGetPropertiesWillExpandPropertyValues()
+    {
+        $this->project->setProperty('foo', 'foo');
+        $this->project->setProperty('bar', '${foo}bar');
+
+        $this->assertEquals(array('foo' => 'foo', 'bar' => 'foobar'), $this->project->getProperties());
```
mpdude: @mrook Surprisingly this does not work ATM. You think it's a glitch or by design?

mrook:
Hm, that's weird, those cases should be handled. I'll investigate!

mpdude:
These tests should pass now

PR #375, opened 2015-01-26
Update for compatibility with newer versions on phpcs

alexbumbacea:


alexbumbacea:
Now the following error is generated: [PHP Error] Argument 3 passed to PHP_CodeSniffer_Reporting::printReport() must be of the type array, string given

PR #376, opened 2015-01-26
update to support phpcs v2.2.*

alexbumbacea:
Fixes error 
[PHP Error] Argument 3 passed to PHP_CodeSniffer_Reporting::printReport() must be of the type array, string given, called in vendor/phing/phing/classes/phing/tasks/ext/PhpCodeSnifferTask.php on line 537 and defined [line 222 of vendor/squizlabs/php_codesniffer/CodeSniffer/Reporting.php]

### DIFF
--- a/classes/phing/tasks/ext/PhpCodeSnifferTask.php
+++ b/classes/phing/tasks/ext/PhpCodeSnifferTask.php
@@ -532,6 +532,9 @@ class PhpCodeSnifferTask extends Task
             $reporting->printReport(
                 $fe->getType(),
                 $this->showSources,
+				array(
+					'colors' => false,
+				),
                 $reportFile,
                 $this->reportWidth
             );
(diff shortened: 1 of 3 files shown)


mrook:
Thanks. Will phpcs 1.5.x still function after this change?

alexbumbacea:
phpcs 1.5.x will not work any more.

Please consider that phpcs 2.2.0 is the latest stable running version also on pear, so i think it will be ok to start using it.

mrook:
I would like to keep 1.5.x working, users may not have upgraded.

throup:
I only discovered this pull request after implementing a similar fix myself. See #379 for a fix which preserves backward compatibility.

mrook:
Closing because of #379, thanks for your efforts @throup and @alexbumbacea !

PR #377, opened 2015-01-27
Making tests somewhat running again...

jvanoostveen:
Before I'm doing some changes, I wanted for a start to be able to run (and succeed) the tests.

I've added tags to certain tests to easily exclude those from running as there is no way to be able to run those without rewriting them. The `rST` tests keep telling me something is not installed after I've installed it and the `svn` tests fail because they can't connect as the issuer is not trusted. Hopefully someday we'll get to bring them under unit tests :).

### DIFF
--- a/src/Diagnostics.php
+++ b/src/Diagnostics.php
@@ -20,6 +20,10 @@
 
 namespace Phing;
 
+use Phing\Io\File;
+use Phing\Io\FileWriter;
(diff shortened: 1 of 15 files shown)


mrook:
Sweet, thanks!

mpdude:
Good job! Adding unit tests is the best help we can get at the moment :)

PR #378, opened 2015-01-29
Simplify UpToDateTask

mpdude:
We don't really want to dispatch an entire subtask here, do we?

The downside is that if you set the target property as a user property (e. g. on the command line), chances are that your user property will win and you don't see the updated value.

But hey, that's how user properties work.

### DIFF
--- a/classes/phing/tasks/system/UpToDateTask.php
+++ b/classes/phing/tasks/system/UpToDateTask.php
@@ -255,11 +255,7 @@ class UpToDateTask extends Task implements Condition
         }
         $upToDate = $this->evaluate();
         if ($upToDate) {
-            $property = $this->project->createTask('property');
-            $property->setName($this->getProperty());
-            $property->setValue($this->getValue());
-            $property->setOverride(true);
-            $property->main(); // execute
+            $this->project->setProperty($this->getProperty(), $this->getValue());
 
             if ($this->mapperElement === null) {
                 $this->log(


siad007:
>The downside is that if you set the target property as a user property (e. g. on the command line), chances are that your user property will win and you don't see the updated value.
>
>But hey, that's how user properties work.

Why not using `Project::setNewProperty()` then?

mpdude:
Doesn't setNewProperty only set it if is not already initialized?

siad007:
It is the same function without the check for user property.

Am Freitag, 30. Januar 2015 schrieb Matthias Pigulla :

mpdude:
No, it will immediately return if the property is already set...?

siad007:
>No, it will immediately return if the property is already set...?

Yes your right. But i think this doesn't matter. Both cases have the same effect, or? I mean, why should we force the property to set if it was set before? But maybe i have overlooked something here. BTW ant is doing a `setNewProperty` as well...

mpdude:
Like this it will be re-set every time, unless it is set as a user property.

User properties are those from the command line and act as "forced" or overridden properties.

siad007:
This makes no sence to me, to re-set every time and having the abillity to overwrite it with a user-property. Or do you have any usage to overwrite this task behavior with a user-property forcing the uptodate task always to be true?

mpdude:
Basically I was just hoping to keep the behaviour as-is :)

The example that comes to mind, however, is that for a set of files you use UpToDate task to figure out whether some compilation must be performed (think of web assets that must be optimized).

You could use a regular property + the UpToDate task to check this.

By overriding the property from the command line, you can force compilation to occur every time or not at all.

Does that make sense?

mrook:
Hm this was closed by renaming 3.0 to future-4.0. Should discuss rebasing this for 3.0.

mrook:
@mpdude this was closed because I just merged 3.0 to master (and deleted the 3.0 branch). Can you rebase this against current master?

PR #379, opened 2015-01-29
Fix PhpCodeSniffer task for version 2.2. (compatible with older versions)

throup:
It involves a crude version check, but this maintains compatibility for older versions of PhpCodeSniffer.


anroots:
Great work @throup! :+1:
Looking forwards to a release, so I can remove `"phing/phing":"dev-master as 2.9.2"`

PR #390, opened 2015-02-01
Project.php minor cleanup

mpdude:
$userProperties actually need not keep the value, just the fact that a particular property is a user property.


mrook:
Rebase?

mpdude:
Did so.

PR #394, opened 2015-02-02
Add exclude option to FileSyncTask.

Polzme:
Add exclude option to rsync command.

### DIFF
--- a/classes/phing/tasks/ext/FileSyncTask.php
+++ b/classes/phing/tasks/ext/FileSyncTask.php
@@ -92,6 +92,12 @@ class FileSyncTask extends Task
     protected $remoteShell;
 
     /**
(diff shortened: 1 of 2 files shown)


mrook:
Thanks for the PR! Can you add a few lines to the user guide (docs/docbook5/en/) for the new attribute?

mpdude:
I might be missing something, but this seems to lack at least a setter method for the new field?

And tests would be nice :)

mrook:
Fuck, totally missed that. I'll add a setter.

mpdude:
A test would have shown it.

To be honest, I don't write tests in real life. Just here for Phing.

mrook:
I usually always write tests, Phing is usually an exception :)

PR #404, opened 2015-02-10
Allow multiple standards for PHP_CodeSniffer

andrewholgate:
This patch has been tested with two, one and no PHP_CodeSniffer standards as arguments.

It is in reference to ticket http://www.phing.info/trac/ticket/1170


mrook:
Thanks @andrewholgate ! Would you mind adapting the documentation as well? (docs/docbook5/en/source)

PR #405, opened 2015-02-10
Allow for fuzzy matches with PHPCPD (Copy/Paste detector)

andrewholgate:
This pull request is in reference to this ticket: http://www.phing.info/trac/ticket/1169


mrook:
Thanks @andrewholgate ! Can you update the docs for this attribute as well?

PR #408, opened 2015-02-16
Solve ticket #1134 and add color support on phar version (all platforms)

llaville:


mrook:
Thanks! Do we need to apply the same logic to bin/phing and/or phing.bat?

llaville:
There is no need to modify the phing.bat, but yes we probably apply the same update on bin/phing.

So perharps a better solution will be to modiy only the bin/phing and not the phing-stub.php (that calls bin/phing)

llaville:
uhm forget to see that the `phing.bat` called directly the `bin/phing.php` and not the `bin/phing`

llaville:
The solution will be to put the modification inside `bin\phing.php` the common script to all enties.
Give me your feedback

phingauto:
Yes that sounds like a good solution. Can you adapt this pr to do that or open a new one?

llaville:
Sure, I can do it, and in a new PR. I can't adapt this one, so I'll close it, and open a new one

llaville:
See new PR #409

PR #415, opened 2015-03-22
Added astro task.

siad007:
Calculates the geocentric apparent longitude in degrees of each planet (+ sun) and rounds to the nearest integer. If any of the planets align, the build will fail with an error message.

This contribution is inspired by https://github.com/lhartikk/AstroBuild


mrook:
I don't think this is a task for general consumption, I'm not going to merge this to master. Thanks though!

PR #426, opened 2015-06-27
Fixed AvailableTaskTest execution under windows.

siad007:
Symlink aware tests will not work on windows.


mrook:
Hmm I thought symbolic links are supported on recent Windows versions? Haven't tested this myself yet.

siad007:
The main problem here is that the shell needs administration previlegs for running the unit tests, which in my opinion is not the best solution.

mrook:
Hmm. Ideally I'd like to avoid that too, but if that's the only way to test if we can create symlinks on Windows..

siad007:
Ah i see, then i will close this issue.

PR #431, opened 2015-07-06
Fixed https://www.phing.info/trac/ticket/1190

siad007:
PropertyRegexTask uses `Regexp()` now.
Fixed https://www.phing.info/trac/ticket/1190

### DIFF
--- a/classes/phing/tasks/ext/property/RegexTask.php
+++ b/classes/phing/tasks/ext/property/RegexTask.php
@@ -84,9 +84,12 @@ class RegexTask extends AbstractPropertySetterTask
         $this->reg = new Regexp();
     }
 
+    /**
+     * @param int $limit
+     */
     public function setLimit($limit)
     {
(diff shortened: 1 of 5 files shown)


mrook:
@siad007 is it possible to retain the limit parameter? That way it won't be a BC breaking change

siad007:
@mrook done

mrook:
Hmm, I guess I should have been more clear, sorry :) The PregEngine class uses preg_replace, which still supports limit. By exposing that in that class, the interface and the RegExp class we can still support the limit parameter from the PropertyRegexTask. Thoughts?

siad007:
No problem - so i would extend the RegExp and PregEngine classes then to support the limit parameter. Would this be an option? I will give it a try in the evening.

mrook:
Yup I think that'd work, thanks

siad007:
@mrook done

PR #437, opened 2015-07-07
Don't apply preg_quote to subject

ip512:
This function add backslashes to characters in string, preventing to use $ in pattern for start by regex.
Additionally it adds \ to version numbers like 1.2.3 (becomes 1\.2\.3)


ip512:
sorry, this is a duplicate of #431

PR #440, opened 2015-07-14
Add debug option for Symfony Console Task

urg:


mrook:
Cool, can you add this new option to the user guide as well?

urg:
I've updated the user guide (hopefully its just the source that needs updating?)

mrook:
This is fine, thanks!

PR #443, opened 2015-07-22
Integrate new NotifySendTask

kenguest:
This will enable phing to send desktop notifications via notify-send.
The code for this was sitting in my separate repo ( https://github.com/kenguest/Phing-NotifySendTask ), but I've added extra logging, unit tests, and documentation.

### DIFF
--- a/classes/phing/tasks/defaults.properties
+++ b/classes/phing/tasks/defaults.properties
@@ -160,6 +160,7 @@ liquibase-rollback=phing.tasks.ext.liquibase.LiquibaseRollbackTask
 liquibase-tag=phing.tasks.ext.liquibase.LiquibaseTagTask
 liquibase-update=phing.tasks.ext.liquibase.LiquibaseUpdateTask
 liquibase=phing.tasks.ext.liquibase.LiquibaseTask
+notifysend=phing.tasks.ext.NotifySendTask
 growlnotify=phing.tasks.ext.GrowlNotifyTask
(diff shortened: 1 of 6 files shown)


siad007:
I think the tests would break on every system where notify-send is not installed.

kenguest:
@siad007 That's why I had added the silent attribute - if it is set to 1 or some other "truthy" value then notify-send will not be executed. 

This is why test/etc/tasks/ext/NotifySendTaskTest.xml looks like this:

 <?xml version="1.0" encoding="UTF-8"?>
<project name="NotifySendTaskTest" default="testEmptyMessage">
  <target name="testEmptyMessage">
    <notifysend silent="1" />
  </target>
</project>

Which I tested by renaming the notify-send binary:
    $ sudo mv /usr/bin/notify-send /usr/bin/notify-send-moved

siad007:
@kenguest hehehe thats clever - i only read your documentation, so i oversee your hidden gem.

kenguest:
@siad007 Thanks :-) It's similar to the idea of using [Operator Presence Controls](https://en.wikipedia.org/wiki/Dead_man%27s_switch) , but I wanted to use a meaningful name for the attribute instead of something vague such as 'opc'. 

I didn't add it to the documentation because I thought most people wouldn't see the point of it ;-)

mrook:
@kenguest can you rebase this PR?

kenguest:
@mrook: done :-)

PR #444, opened 2015-08-04
Updated JsHintTask to work with latest JSHint

mhujer:


mrook:
@mhujer thanks! Could you please amend the user guide to state what version of jshint we support?

mhujer:
@mrook I've updated the docbook docs, you are rebuilding them before release, right?

mrook:
Yes, I will. Thanks!

PR #447, opened 2015-08-11
Adding support for multi-character beginToken and endToken when calli…

hed854:
…ng ReplaceTokens with *filterreader* tag


mrook:
Thanks!

PR #451, opened 2015-08-17
append RenameTask

alzasr:
Append RenameTask. This task rename file or folder. MoveTask use method copy and set new mode for files. 

Sorry for my English. I'm from Russia.

### DIFF
--- a/classes/phing/tasks/defaults.properties
+++ b/classes/phing/tasks/defaults.properties
@@ -36,6 +36,7 @@ property=phing.tasks.system.PropertyTask
 propertyprompt=phing.tasks.system.PropertyPromptTask
 record=phing.tasks.system.RecorderTask
 reflexive=phing.tasks.system.ReflexiveTask
+rename=phing.tasks.system.RenameTask
 resolvepath=phing.tasks.system.ResolvePathTask
 retry=phing.tasks.system.Retry
 taskdef=phing.tasks.system.TaskdefTask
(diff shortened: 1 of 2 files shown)


siad007:
What would be the difference to simply use `move` instead?
```
    <move file="/file/A" tofile="/file/B" />
```

alzasr:
MoveTask set new mode (with umask) for all files and directory. if my files had the necessary mode, then after MoveTask i need use Chmod for set needen mode. RemaneTask saves old mode. Move = Copy + Delete, if size of the directory is large, it will take a lot of time. Rename much faster.

Sorry for my English. I'm from Russia.

mrook:
I think it'd make more sense to extend/adapt the move task so it tries to rename complete dirs, and if that doesn't work, revert to the current copy+delete strategy.

alzasr:
For me the bigger problem was the change mode. I spent half an hour to understand why my project is not working after MoveTask. I think it does not solve the selection strategy. Here need a new option - save old mode or set new. IMHO, RenameTask more understandable than MoveTask with option.

Sorry for my English. I'm from Russia.

mrook:
Your english is fine :) If we actually try a rename first, then the mode does not change. I'm specifically looking at https://github.com/apache/ant/blob/master/src/main/org/apache/tools/ant/taskdefs/Move.java#L114 for inspiration.

alzasr:
Current behavior MoveTask is a bug and not is a feature? Need to fix this bug?

mrook:
Not really a bug, it was just never implemented.

mrook:
I'm closing this PR in favor of adding actual rename functionality to the move task at some point.

PR #458, opened 2015-08-27
Added FlexInteger helper.

siad007:
Added ant like `FlexInteger` helper class, which can be used for phing task attribute setter methods to allow the build file to specify an integer in either decimal, octal, or hexadecimal format.


mrook:
I'm not entirely sure about how useful this class is, as it seems nothing more than a wrapper around intval(). What are your thoughts?

siad007:
@mrook you are so right it makes no sense in php context.

PR #459, opened 2015-08-27
ApplyTask - added support for dirsets.

siad007:
ApplyTask with support for nested dirset elements.


mrook:
A small note in the user guide would be nice.

siad007:
@mrook done ;)

PR #460, opened 2015-08-28
Fixing bug in Yaml file parser.

grasmash:
Looks like the most recent modification to the Yaml feature introduced a bug. This should fix it. Also need to see why regression test did not catch this.

### DIFF
--- a/classes/phing/system/io/YamlFileParser.php
+++ b/classes/phing/system/io/YamlFileParser.php
@@ -40,8 +40,9 @@ class YamlFileParser implements FileParserInterface
         try {
             // We load the Yaml class without the use of namespaces to prevent
             // parse errors in PHP 5.2.
-            $parser = '\Symfony\Component\Yaml\Parser';
-            $properties = $parser->parse($file->getAbsolutePath());
+            $parserClass = '\Symfony\Component\Yaml\Parser';
+            $parser = new $parserClass;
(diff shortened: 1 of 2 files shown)


grasmash:
@mrook This commit should fix the immediate problem.

However, the matter of the regression tests passing is curious to me. It seems that many of the regression tests are actually skipped in Travis:

````
  [phpunit] testParseFileFileNotReadable SKIPPED
  [phpunit] testParseFileFileIncorrectYaml SKIPPED
  [phpunit] testParseFileFile SKIPPED
````

Do you know why this is?

**Comment by mrook** [comment]
I think the tests are skipped because 

```
if (phpversion() < 5.3 || !class_exists('Yaml')) {
```

in YamlFileParserTest.php

I think I read over this line, is the extension really required?

grasmash:
Yes, it's totally necessary. Let me see if adding the namespace will cause the test to run.

grasmash:
@mrook Looks like it's working now. Tests are both running and passing.

siad007:
@mrook i think the test should be marked as skipped and not incomplete because the test itself is complete but the assumptions are not met.

mrook:
@siad007 yup, you're right, thanks. Fixed in https://github.com/phingofficial/phing/commit/1d7987eb342cd6d3dde7aff2cb81e6e6569b7695

PR #463, opened 2015-09-21
Add propertyName to PhpCodeSniffer

urg:
We had a case in the past where a sniff failed to run in our setup.  By having the sniffs in a property we would be able to validate that the sniffs were in fact used.


mrook:
Useful PR, thanks!

PR #467, opened 2015-10-27
Implement solution for ticket 1222

llaville:
Take care that depending of your choice, some code may be reduced if you consider to refactor base class BatchTest and changed private properties and methods to protected


mrook:
Thanks for the PR, but Phing needs to remain compatible with at least 5.3 (and preferably 5.2 as long as we can get away with it). "yield" is 5.5+

llaville:
@mrook Yes, I understand, but if you look deep inside, you 'll see that only platform with PHP 5.5+ will used the generator feature. PHP 5.2, 5.3, 5.4 will still use default behavior.

llaville:
@mrook Did you have time to deepthly explored my solution and see that it is backward compatible with all versions, especially lower 5.4, 5.3, 5.2

mrook:
No, I'll try to take a look at it soon. Some tests are failing though.

llaville:
ping @mrook any news since 2 months ?

mrook:
Can you rebase the PR and see if the failing tests can be fixed?

PR #472, opened 2015-11-09
Set ATTR_EMULATE_PREPARES => true for PgSQL connections

valinurovam:
Hi all.
It is very important to set PDO::ATTR_EMULATE_PREPARES => true. Because some projects work with pgsql through pgBouncer (more information here https://pgbouncer.github.io/faq.html), and without this set we have an error like this

```
Previous exception 'PDOException' with message 'SQLSTATE[26000]: Invalid sql statement name: 7 ERROR:  prepared statement "pdo_stmt_00000001" does not exist' in /usr/share/php/phing/tasks/ext/dbdeploy/DbDeployTask.php:181
```

With this set attribute all connections through pgBouncer or direct to pgsql is work.

Thank you.

### DIFF
--- a/classes/phing/tasks/ext/dbdeploy/DbmsSyntaxPgSQL.php
+++ b/classes/phing/tasks/ext/dbdeploy/DbmsSyntaxPgSQL.php
@@ -29,6 +29,14 @@
 class DbmsSyntaxPgSQL extends DbmsSyntax
 {
     /**
+     * @param \PDO $db
+     */
+    public function applyAttributes($db)
+    {
+        $db->setAttribute(\PDO::ATTR_EMULATE_PREPARES, true);
+    }
+
+    /**
      * @return string
      */
     public function generateTimestamp()


mrook:
I have no experience with postgres, but it seems that forcing emulation of prepared statements seems like a bit of a brute-force solution. Can you think of a different approach?

valinurovam:
Yes, my current solution is inheritance of DbDeployTask and declare new one in deploy.xml like taskdef
https://gist.github.com/valinurovam/be9425b1730731f8766d

But search on common solution, cause current looks like "crutch"

mrook:
What do you mean with "crutch"?

valinurovam:
I mean dognail.

Sorry, bad english.

mrook:
I'm still not sure what you're trying to say :)

valinurovam:
May be "kludge" is better word)

mrook:
How about making this an optional attribute in DbDeployTask? With the right documentation ofcourse, so users know how to enable it.

mrook:
Closing for lack of feedback.

PR #480, opened 2015-12-30
Export ignore docs file

fezfez:
The Docs folder, once downloaded is 5.64 MB, which is large when it is not a necessity


fezfez:
Any news ?

PR #481, opened 2016-01-03
Add foreachparallel-task to phing

mkrauser:
Hi guys,

since phing has a ParallelTask and a ForeachTask, it makes sense to have a ForeachParallelTask too. I've developed a standalone-library for this a while ago (https://github.com/mkrauser/phing-foreach-parallel-task). 
I would love to see it in Phing-Core. Since the new Task inherits from ForeachTask, I had to make some changes to it as well to make it more open for inheritance.

Thanks for the awesome work!

### DIFF
--- a/classes/phing/tasks/system/ForeachTask.php
+++ b/classes/phing/tasks/system/ForeachTask.php
@@ -34,17 +34,25 @@ include_once 'phing/tasks/system/PhingTask.php';
  * <foreach list="values" target="targ" param="name" delimiter="|" />
  *
  * Attributes:
- * list      --> The list of values to process, with the delimiter character,
- *               indicated by the "delimiter" attribute, separating each value.
- * target    --> The target to call for each token, passing the token as the
- *               parameter with the name indicated by the "param" attribute.
- * param     --> The name of the parameter to pass the tokens in as to the
- *               target.
- * delimiter --> The delimiter string that separates the values in the "list"
- *               parameter.  The default is ",".
+ * list        --> The list of values to process, with the delimiter character,
+ *                 indicated by the "delimiter" attribute, separating each value.
+ * target      --> The target to call for each token, passing the token as the
+ *                 parameter with the name indicated by the "param" attribute.
+ * param       --> The name of the parameter to pass the tokens in as to the
+ *                 target.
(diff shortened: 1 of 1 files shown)


mrook:
Hm, I like the idea very much, but wouldn't it be more interesting to see if the parallel task can be refactored (extract a few classes, etc.) and add a "parallel=true|false" option to the foreach task?

mkrauser:
@mrook I thought about that as well, but I did not want to touch the foreach-task before talking to core-devs. If you like, I can work on that

mrook:
That would be great!

mkrauser:
@mrook I've updated the pull request to implement concurrency in foreach-task. 

There are two things that bother me:
1. First, even if the tasks are not run in parallel, a new callee is created for every iteration. I can change that, but the code will become a little unreadable
2. The parallelManager is passed to process(...) as parameter. Personally I would rather implement this as class-property.

Please take a look an let me know your thoughts.

mrook:
Thanks! I'll have a look soon (pretty busy at the moment).

mrook:
Hi @mkrauser sorry it took me so long to review/reply. I like the general gist of it, but I have a few minor points:

- there is some code duplication, this block is repeated and can probably be re-used by extracting it to a method:
                if($this->parallel === true) {
                    $worker = new DocBlox_Parallel_Worker(
                        array($callee, 'main'),
                        array($callee)
                    );
                    $parallelManager->addWorker($worker);
                } else {
                    $callee->main();
                }
- parallelManager can be a class property (as you stated yourself)
- should createCallee be public?
- I'm not against marking more properties as protected, but you add a few private properties in your commit; what should we do here?

mkrauser:
Hey @mrook, 
it also took me a while, I was pretty busy with other projects as well.

I've update the PR to reflect your suggestions. There's only one thing that I'm not entirely happy with: the name of configureCallee(...), somehow It does not feel right. 
If you have a suggestion for a better name or any other improvements, please get back to me.

Thanks for your review!

mrook:
Thanks for your contribution! Closing this because of inactivity, please re-open or submit a new pull request if you have time to fix the mentioned issues.

PR #486, opened 2016-02-14
Exclude multiple files/directories

shanavas786:
Use comma separated values to exclude multiple files/directories.
e.g.: exclude="a,b" will produce command lineoptions --exclude="a" --exclude="b"`


mrook:
Thanks!

PR #488, opened 2016-03-15
Fixing -propertyfile option and adding yaml support for it.

grasmash:
I found that passing `-propertyfile <filename.yml>` was not setting the properties values as expected. The first reason for this was that the command option assumes that the file will use the IniParser. This has been addressed and did not support .yml files.

However, after adding yaml support I found that the expected property values still were not set. I'm not sure why. The passed `-propertyfile` was not overriding existing property values--merely defining new ones. Regardless, adding them to the $definedProps array seems to have solved it.

For background, I was using the `-f build.xml` option as well as loading various property files via the `<property file="<filename>" />` method. I'm not sure if this could be related.

### DIFF
--- a/classes/phing/Phing.php
+++ b/classes/phing/Phing.php
@@ -127,6 +127,9 @@ class Phing
     /** Whether to capture PHP errors to buffer. */
     private static $phpErrorCapture = false;
 
+    /** Whether to values in a property file should override existing values. */
+    private $propertyFileOverride = false;
+
     /** Array of captured PHP errors */
     private static $capturedPhpErrors = array();
 
@@ -419,6 +422,11 @@ class Phing
             unset($args[$key]);
         }
 
+        if (false !== ($key = array_search('-propertyfileoverride', $args, true))) {
+            $this->propertyFileOverride = true;
+            unset($args[$key]);
+        }
+
         // 3) Finally, cycle through to parse remaining args
         //
         $keys = array_keys($args); // Use keys and iterate to max(keys) since there may be some gaps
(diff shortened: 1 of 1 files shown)


grasmash:
I found that passing `-propertyfile <filename.yml>` was not setting the properties values as expected. The first reason for this was that the command option assumes that the file will use the IniParser. This has been addressed and did not support .yml files.

However, after adding yaml support I found that the expected property values still were not set. I'm not sure why. The passed `-propertyfile` was not overriding existing property values--merely defining new ones. Regardless, adding them to the $definedProps array seems to have solved it.

For background, I was using the `-f build.xml` option as well as loading various property files via the `<property file="<filename>" />` method. I'm not sure if this could be related.

grasmash:
@mrook Does this pull request require any changes before being merged?

mrook:
Overriding existing property files sounds like a backwards-incompatible change, so I'm not sure that's something we can merge. The fixes look fine.

grasmash:
What if we introduced an additional option to specify whether the build file should override defined properties? Would that suffice to make backwards compatible?

E.g., `phing -propertyfile filename.yml -propertyfileoverride

mrook:
That could work :)

grasmash:
The errors in the Travis output seem related to an SVN checkout issue, which seems unrelated to changes made in this PR:
```
Server sent unexpected return value (429 Too Many Requests) 
```

@mrook I've introduce the new -propertyfileoverride option as discussed. Please let me know if any further changes are required.

CashWilliams:
Looks good to me. I'm running this branch for some builds and working well.

mrook:
Looks good. One last thing: can you clean up the indentation (should be 4 spaces per level of indentation) and squash the commits?

grasmash:
@mrook Updated and squashed.

mrook:
Thank you!

PR #492, opened 2016-03-31
Fix Windows encoding convertion

mizunashi-mana:
I got an error on run phing.

```
> phing prepare
Notice: iconv(): Wrong charset, conversion from `UTF-8' to `Windows-932//IGNORE' is not allowed in vendor\phing\phing\classes\phing\system\io\Win32FileSystem.php on line 635
Buildfile:  does not exist!
```

Now, phing use `iconv` or `mb_convert_encoding` for fixing windows encoding.
However, `Windows-${codepage}` is not usual codepage encoding format.

My `cmd.exe` use codepage 932.
`CP932` exists in supported encoding for `iconv` and `mbstring`, but `Windows-932` is not.
So, `fixEncoding` function failed and returned not string...

I request to use `CP` for encoding prefix.
Thanks!

## Refs

 * https://www.gnu.org/software/libiconv/
 * http://php.net/manual/ja/mbstring.supported-encodings.php

### DIFF
--- a/classes/phing/system/io/Win32FileSystem.php
+++ b/classes/phing/system/io/Win32FileSystem.php
@@ -630,7 +630,7 @@ class Win32FileSystem extends FileSystem
      */
     private function fixEncoding($strPath)
     {
-        $codepage = 'Windows-' . trim(strstr(setlocale(LC_CTYPE, 0), '.'), '.');
+        $codepage = 'CP' . trim(strstr(setlocale(LC_CTYPE, 0), '.'), '.');
         if (function_exists('iconv')) {
             $strPath = iconv('UTF-8', $codepage . '//IGNORE', $strPath);
         } elseif (function_exists('mb_convert_encoding')) {
--- a/classes/phing/system/io/Win32FileSystem.php
+++ b/classes/phing/system/io/Win32FileSystem.php
@@ -630,7 +630,7 @@ class Win32FileSystem extends FileSystem
      */
     private function fixEncoding($strPath)
     {
-        $codepage = 'CP' . trim(strstr(setlocale(LC_CTYPE, 0), '.'), '.');
+        $codepage = 'CP' . trim(strstr(setlocale(LC_CTYPE, ''), '.'), '.');
         if (function_exists('iconv')) {
             $strPath = iconv('UTF-8', $codepage . '//IGNORE', $strPath);
         } elseif (function_exists('mb_convert_encoding')) {


mrook:
Changing "Windows-" to "CP-" will probably cause a regression in https://github.com/phingofficial/phing/commit/84e4050e06c678ff079f11d07bf2ed03c7fa35d5 / https://www.phing.info/trac/ticket/1198

Is there some other way to determine available codepages? I don't do much Windows development, and I don't have any special codepages set on my Windows machine.

mizunashi-mana:
@mrook Thank you for feedback!

I try on PHP 7.4 (on psysh v0.7.2):

```php
iconv('UTF-8', 'Windows-1251//IGNORE', 'something'); // success
iconv('UTF-8', 'CP1251//IGNORE', 'something'); // success
```

If iconv receive not support encoding, return `false` by ignore.

```php
mb_convert_encoding('something', 'Windows-1251', 'UTF-8'); // success
mb_convert_encoding('something', 'CP1251', 'UTF-8'); // success
```

If mb_convert_encoding receive not support encoding, throw error.

It seems to be no problem.
On https://gist.github.com/hakre/4188459 (iconv encodings) and http://php.net/manual/ja/mbstring.supported-encodings.php (mbstring encodings), `CP-` is more supported than `Windows-` and there are `CP-` aliases for all of `Windows-` encoding.
If we trust documents on the net, this request will be fine on Windows ;)

My Japanese Windows is using `CP932`.  This is usual code page in Japan.  However, `CP932` doesn't have `Windows-` alias unfortunately...

classes/phing/system/io/Win32FileSystem.php
```diff
@@ -630,7 +630,7 @@ public function lister($f)
      */
     private function fixEncoding($strPath)
     {
-        $codepage = 'Windows-' . trim(strstr(setlocale(LC_CTYPE, 0), '.'), '.');
+        $codepage = 'CP' . trim(strstr(setlocale(LC_CTYPE, 0), '.'), '.');
```
siad007: As shown in the iconv list of encodings https://gist.github.com/hakre/4188459 `Windows-` is the alias for `CP` as mentioned by @mizunashi-mana, there will be no regression to this. But i would suggest changeing `setlocale(LC_CTYPE, 0)` to `setlocale(LC_CTYPE, '')` to get allways the default locale under windows and not an `C` if not defined in settings properly.

mrook:
Thanks for the updates! @mizunashi-mana, can you update your pull request with @siad007 's comments? I'll then merge it.

mizunashi-mana:
@mrook @siad007 I updated this request!

PR #494, opened 2016-04-16
DBDeploy script update to address an issue that arises when following a popular tutorial

andrewperry:
Adding a warning if user includes --// in a DB delta as it will likely cause a schema migration to fail and this is a common issue due to a popular tutorial using '--//' instead of '-- //' for comments

### DIFF
--- a/classes/phing/tasks/ext/dbdeploy/DbDeployTask.php
+++ b/classes/phing/tasks/ext/dbdeploy/DbDeployTask.php
@@ -255,6 +255,12 @@ class DbDeployTask extends Task
                 $fullFileName = $this->dir . '/' . $fileName;
                 $fh = fopen($fullFileName, 'r');
                 $contents = fread($fh, filesize($fullFileName));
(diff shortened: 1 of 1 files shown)


mrook:
Thanks! One small item: instead of echo-ing to console, could you use the Phing logger with an appropriate level?

andrewperry:
Thanks, I expected there would be a better way but thought I'd at least contribute where I'd got up to. 

I'll look into it.

andrewperry:
How's that @mrook ?

mrook:
As it's a warning, I think $this->log(..., Project::MSG_WARN) would be more appropriate, in which case the "Warning: " part of the string is no longer necessary.

andrewperry:
Thanks @mrook - updated patch pushed.

mrook:
Excellent. One final nag, could you squash those commits to a single one? I'll then merge it :)

andrewperry:
Thanks @mrook

PR #501, opened 2016-05-10
Add set of tasks for working with Mercurial.

kenguest:
Phing doesn't currently support working with Mercurial. This pull request adds such functionality, including documentation and related unit tests. It utilises @siad007 's versioncontrol_hg library.


mrook:
I'm not entirely sure why the builds fail, can you take a look?

kenguest:
@mrook I found out why the build was failing in that rather spectacular manner - when pulling in HgBase.php, I needed to refer to it as 'phing/tasks/ext/hg/HgBaseTask.php', not just 'HgBaseTask.php' and I also needed to address a composer dependency.

It works now on all versions apart from 5.2 and 5.3 - the versioncontrol_hg package by @siad007 which I am using requires 5.4 or higher...

siad007:
Would it be of any help to make an Release of the Package to support php <
5.4 ?

Am Freitag, 13. Mai 2016 schrieb Ken Guest :

kenguest:
@siad007 perhaps, though to be honest I'm not sure if it'd be worth the effort. I mean 5.2, 5.3. and 5.4 aren't "supported versions" any more - http://php.net/eol.php.
@mrook what do you think? Should we be worrying about supporting such old versions of PHP any more? I know it's a bit far reaching, but should we perhaps just set 5.3 as an allowable failure in the .travis.yml file?

mrook:
5.2 is an allowed failure, but I want Phing to still run on that platform. Same goes for 5.3. The HG tasks don't have to work on those platforms, but Phing should boot & run.

kenguest:
@mrook, do you want to check this over again? Copious use of version_compare(PHP_VERSION....) and such, including a tweak of the composer.lock file seem to have this working ok now.

mrook:
It looks pretty good, with one small nag: "use" can't be used I'm afraid, as that breaks BC PHP 5.2. I'm trying not to do that in the 2.x branch.

siad007:
Instead of using `use` we could check and load the factory with the help of something like this:
```php
    /**
     * Load factory.
     *
     * @return void
     *
     * @throws BuildException
     */
    private function loadFactory()
    {
        if (version_compare(PHP_VERSION, '5.4', '<')) {
            throw new BuildException("MercurialTask requires at least PHP 5.4 installed.");
        }
        @include_once 'vendor/autoload.php';
        if (!class_exists('\\Siad007\\VersionControl\\HG\\Factory')) {
            throw new BuildException("Mercurial Task requires VersionControl\\HG to be installed");
        }
    }
    /**
     * Get the factory instance.
     *
     * @return \Siad007\VersionControl\HG\Factory
     */
    private function getFactoryInstance()
    {
        if (self::$factory === null) {
            $vchq = '\\Siad007\\VersionControl\\HG\\Factory';
            self::$factory = new $vchq;
        }
        return self::$factory;
    }
```

mrook:
A factory would be fine yes.

kenguest:
That's done now and is passing for 5.2 in relation to linting the files with a 5.2 version of php.

PR #505, opened 2016-06-02
Removes executable bit from...

donatj:
I on occasion run a security sweep on our codebase. One of the things it indicates are things that are executable that probably shouldn't be, eg: non-binaries that don't start with `#!`

Our `vendor/phing` directory lit up like a christmas tree.

I wrote a little script that removed the exec bit on everything it suggested.


donatj:
The travis failure was a network timeout, and not the change.

mrook:
Thanks!

PR #520, opened 2016-09-07
Make SassTask compatible with PHP 7

pfrenssen:
The new SassTask is not compatible with PHP 7. Interestingly, earlier on in the development of the PR it was compatible, but at some point backwards compatibility with PHP 5.2 was added but this broke PHP 7.

This error occurs when trying to use the SassTask with scssphp on PHP 7:

> call_user_func_array() expects parameter 1 to be a valid callback, non-static method Leafo\ScssPhp\Compiler::__construct() cannot be called statically [line 1219 of /home/pieter/v/joinup/vendor/phing/phing/classes/phing/tasks/ext/SassTask.php]

### DIFF
--- a/classes/phing/tasks/ext/SassTask.php
+++ b/classes/phing/tasks/ext/SassTask.php
@@ -1213,17 +1213,10 @@ class SassTask extends Task
      */
     public function getNewCompiler()
     {
-        // Support for namespaces was added in PHP 5.3. Provide a fallback for
-        // PHP 5.2.
-        if (version_compare(PHP_VERSION, '5.3.0') < 0) {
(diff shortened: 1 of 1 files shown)


mrook:
The problem is that "return new Leafo\ScssPhp\Compiler();" is not a parse-able line under 5.2...

siad007:
I think the following snipped would be enough?

```php
public function getNewCompiler()
{
    $compiler = '\\Leafo\\ScssPhp\\Compiler';
    return new $compiler;
}
```

mrook:
Yep, it would.

pfrenssen:
I've updated this on my phone. I've not had the chance to test this but I believe too that this works fine on all PHP versions.

PR #521, opened 2016-09-09
Displaying filename in CopyTask when only one file is copied.

grasmash:


grasmash:
Presently, if the CopyTask is used to copy one file, it outputs `Copying 1 file to [destination]", which isn't very informative. If only one is copied, we mine as well display the file name. When more than one is copied, displaying the number of files is fine.

With this change we're not increasing the number of lines of output, only making the output more relevant.

PR #525, opened 2016-09-11
Added new function to append task.

siad007:
Append task with new attributes:
- append
- overwrite
- fixlastline
- eol

New nested elements:
- header
- footer
- path


mrook:
Could you also add the new attributes/elements to the grammar file (in etc). Forgot to ask that for the sleep task, but I'll add it there.

siad007:
:)

mrook:
Cheers!

PR #527, opened 2016-09-16
Added runtarget task.

siad007:
Phing task that runs a target:

``` xml
  <target name="run">
    <runtarget target="anotherTarget" />
  </target>

  <target name="anotherTarget">
    <echo msg="THIS COMES FROM ANOTHER TARGET"/>
  </target>
```

### DIFF
--- a/classes/phing/tasks/defaults.properties
+++ b/classes/phing/tasks/defaults.properties
@@ -40,6 +40,7 @@ record=phing.tasks.system.RecorderTask
 reflexive=phing.tasks.system.ReflexiveTask
 resolvepath=phing.tasks.system.ResolvePathTask
 retry=phing.tasks.system.Retry
+runtarget=phing.tasks.system.RunTargetTask
 sleep=phing.tasks.system.SleepTask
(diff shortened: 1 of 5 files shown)


grasmash:
How does this differ from phingcall?

siad007:
The main difference is the scope, which means that any property it sets will not be visible to the build that used `<phingcall>`. `<runtarget>` does not create a new Project, so all properties set are visible to the "caller". So there would be an performance boost if we take `<runtask target="foo.bar"/>` instead of `<phingcall inheritrefs="true"/>` but in fact it would be cleaner/saver to use the build in phingcall, that's why i moved this to the ext/optional tasks.

siad007:
Related to #576 - reminder to myself: needs rebase

mrook:
@siad007 I'd like to go ahead and merge this, but we probably need to include something in the documentation to warn users about the potential pitfalls/dangers of using ```runtarget``` vs ```phingcall```, what do you think?

siad007:
@mrook I will add a note in the documentation.

PR #530, opened 2016-10-04
Fix missing include to FileParserFactory 

llecaque:
When using directly the phar without composer autoload with the -propertyfile option, I was having a FATAL ERROR
$ php phing.phar -propertyfile config.ini help 
PHP Fatal error:  Class 'FileParserFactory' not found in phar:///home/workspace/phing.phar/classes/phing/Phing.php on line 515

It seems that in PR #488 , the include for class FileParserFactory was missing


mrook:
Thanks!

PR #590, opened 2016-10-10
Adding 0 and 1 strings as true and false values in StringHelper.

grasmash:


grasmash:
I find that people commonly pass "0" and "1" as strings for boolean properties. These should be evaluated as false and true.

grasmash:
@mrook I believe this is a false test failure. I this PR acceptable? It seems like a minor, backwards compatible change.

mrook:
@grasmash can you rebase?

mrook:
Also, I'm trying to decide whether this is a BC break or not ;-)

grasmash:
Rebased!

siad007:
Related to #261

mpdude:
:+1: for this change

mrook:
@grasmash I think this would mean a BC break. Shouldn't be that much of a problem, I'm planning to release 3.0 early next year.  Can you rebase this PR against the 3.0 branch?

PR #599, opened 2016-10-19
Improving error message for HttpCondition.

grasmash:


grasmash:
The existing error message for a failed curl_exec() in HttpCondition is not quite accurate. While true that it may indicate a malformed URL, the failure always indicates that no response was received. Given that the URL is printed to screen, the user should be able to determine if it is malformed on their own.

mrook:
@grasmash thanks, can you rebase this?

grasmash:
Rebased!

PR #617, opened 2016-11-04
Fix formated time

sui77:
Rounding error results in DefaultLogger::formatTime(90) => "2 minutes 30.00 seconds" but should be "1 minute 30.00 seconds" instead.


mrook:
Thanks!

PR #620, opened 2016-11-07
Fixed notice "A non well formed numeric value encountered" in PHP 7.1

kukulich:


lastzero:
What release will this be in? It's now 3 months later and I still see this error when using the latest ~2.0 via composer.

mrook:
It's in 2.16, which was released Dec 22.

lastzero:
Seems like an issue with composer dependencies on my side... I require symfony/symfony ^2.8|^3.0 and that seems to cause issues with 2.16... composer only installs Phing 2.15.2 currently... I'm investigating it!

lastzero:
For a reason unknown to me, 2.16 requires `symfony/yaml` in version `^3.1` while your master requires

```
"symfony/console": "^2.8|^3.0",
"symfony/yaml": "^2.8|^3.1"
```

That is compatible with Silex 1 projects. Silex 1 does not support Symfony ^3.1 so Phing version `2.15.2` will be installed and not `2.16.0`. Silex 2 now supports Symfony 3.x, so Phing 2.16.0 can be installed. Maybe this helps, if somebody has the same issue.

mrook:
If required I can roll a 2.16.1 with `symfony/yaml` set to `^2.8|^3.1`.

lastzero:
Sounds good. Why not?

mrook:
Because we're working on 3.0 and rolling a release requires time :) Also, there are no other things that need patching in 2.16.

lastzero:
I assume there are other users with the same issue, so it might help. Since I just did the Silex upgrade anyways, I don't need it anymore :)

mrook:
2.16.1 was released earlier this month: https://github.com/phingofficial/phing/releases/2.16.1

XedinUnknown:
`2.16.1` requires YAML `^3.1`, which requires PHP 5.5.

mrook:
Ah, I wasn't aware of that. I don't think we've ever targeted symfony/yaml 2.8 before, which means we've been incompatible with < 5.5 for a while. Alas, that compatibility was hard to maintain (and Phing 2.x is deprecating anyway).

I'll try to roll a 2.16.2 somewhere next week, with this fixed.

mrook:
Continuing in #919.

PR #621, opened 2016-11-10
[TECH] Change the way to list the sql patches

codisart:


mrook:
Hi @codisart, what are your thoughts behind this PR? This breaks compatibility with existing build scripts and the task documentation, and there's no way to override the ```listFile``` value.

codisart:
Hi, @mrook,

First of all, I'm sorry to open this PR without much explanations.
I was not on a correctly configured computer when I tried this and I thought that it could be a good idea to use travis for TDD.

We work with phing for almost every project in our company and specifically with dbdeploy for our databases migrations.
And one peculiar project, we are a team of 20 peoples and with several version at the same time. We encountered several times problems in deployment of several missing scripts. To be true it happens even more often on ou developpment environnements. 

I was pretty frustated

mrook:
Hi @codisart, instead of a version number you could prefix your migration files with a timestamp. It's unlikely that two developers will start writing a migration at exactly the same time (down to the second), so that should allow some breathing room.

mrook:
Closing this PR because of the above comments.

PR #635, opened 2016-12-22
Namespacing

cdburgess:
- adding the Phing namespace to ALL php files in the commit


mrook:
Hi @cdburgess, thanks for your work! However, 2.x (current master) should remain 5.2+ compatible.

The new major version (https://github.com/phingofficial/phing/tree/3.0, which will be released sometime early next year) will be PHP 5.6+. Because I don't want to wait too long with pushing that new version, I'm not sure namespacing that code is the right way to go, I'd like to take a little bit more time to come up with a solid structure / interface / API. Some namespacing work has been done on a "future" 4.x branch (https://github.com/phingofficial/phing/tree/future-4.0)

PR #645, opened 2017-01-09
Fixes #644: Change default verbosity of replaceProperties() to MSG_DEBUG.

### LINKED ISSUE #644: Duplicate log output from replaceProperties()
I find that importing the following build properties:

```
drush:
  cmd: ${drush.bin} @${drush.alias} -r ${docroot} -l ${multisite.name}
```

Will output the following when the `-verbose` argument is used:

```
Property ${drush.alias} has not been set.
Property ${drush.alias} has not been set.
Property ${drush.alias} has not been set.
Property ${drush.alias} has not been set.
Property ${drush.alias} has not been set.
```

This seems like unnecessary noise. I only need to be notified once about a property value, unless it changes.

In practice, this prevents me from using `-verbose` because it is too noisy. But, I still need a setting that is more verbose than 'info'.

I propose that as a low-effort solution, we change this output to the MSG_DEBUG level. In the long term, each property's value should be displayed only once unless it is changed.

grasmash:


siad007:
Hi @grasmash can you please recheck, if it is still an issue for you? I think the function was moved to PropertyHelper::replaceProperties() and works a bit different now.

siad007:
I will close this PR as of changes to the core - feel free to reopen if required.

PR #648, opened 2017-01-11
Moved ResolvePathTask to PropertyTask.

siad007:
Moved function of
`<resolvepath propertyName="resolved.dir" dir="."/>`
to
`<property name="resolved.dir" location="."/>`
and deprecated the use of it.

### DIFF
--- a/classes/phing/IntrospectionHelper.php
+++ b/classes/phing/IntrospectionHelper.php
@@ -145,7 +145,7 @@ class IntrospectionHelper
                 // There are a few "reserved" names that might look like attribute setters
                 // but should actually just be skipped.  (Note: this means you can't ever
                 // have an attribute named "location" or "tasktype" or a nested element named "task".)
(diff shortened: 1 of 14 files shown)


mrook:
I like this! Maybe add ```setLocation``` to increase compatibility with Ant?

siad007:
Would be awsome but unfortunaly setLocation is a forbidden word. see IntrospectionHelper:
```php
    // There are a few "reserved" names that might look like attribute setters
    // but should actually just be skipped.  (Note: this means you can't ever
    // have an attribute named "location" or "tasktype" or a nested element named "task".)
    if ($name === "setlocation" || $name === "settasktype" || $name === "addtask") {
        continue;
    }
```

mrook:
Good point, may have to think about that.

mrook:
Yeah we can't actually get rid of that. ```setLocation``` is already defined in ```ProjectComponent.php```. This works in Ant because of method overloading.

siad007:
@mrook this would work - what do you think?

Issue #584, opened 2016-10-09
PHPUnitReportTask fails with XSLTProcessor::importStylesheet() unable to read phar:/usr/local/bin/phing/etc/str.replace.function.xsl (Trac #1240)


phing-issues-importer:
Hi,

my build fails with:

XSLTProcessor::importStylesheet(): I/O warning : failed to load external entity "phar:/usr/local/bin/phing/etc/str.replace.function.xsl", file: phar:///usr/local/bin/phing/classes/phing/tasks/ext/phpunit/PHPUnitReportTask.php

See buildfile below: It successfully executed the tests and creates a valid logfile.xml as expected. Converting test results with <<phpunitreport> then fails.

Using both Phing and PHPUnit PHAR archives.

After extracing the Phing PHAR, I see that is DOES contain phar:/usr/local/bin/phing/etc/str.replace.function.xsl.

Environment:
Phing 2.13.0
PHP 5.6.14-0+deb8u1 (cli) (built: Oct  4 2015 16:13:10)
libxslt Version 1.1.28 with XSL enabled
Registered PHP Streams: https, ftps, compress.zlib, compress.bzip2, php, file, glob, data, http, ftp, phar, zip
PHPUnit 4.8.7 by Sebastian Bergmann and contributors.
Linux version 3.16.0-4-amd64 (debian-kernel@lists.debian.org) (gcc version 4.8.4 (Debian 4.8.4-1) ) https://www.phing.info/trac/ticket/1 SMP Debian 3.16.7-ckt11-1+deb8u3 (2015-08-04)

Buildfile (paths changed, please adjust as needed):

```
<?xml version="1.0"?>

<project name="MyProject" basedir="." default="tests_with_report">
  <target name="prepare">
    <mkdir dir="testreport"/>
  </target>

  <target name="clean">
    <delete dir="testreport"/>
  </target>

  <target name="tests_with_report" depends="prepare">
    <phpunit bootstrap="bootstrap.php" pharlocation="/usr/local/bin/phpunit">
      <batchtest>
        <fileset dir="tests">
          <include name="config/*Test.php"/>
        </fileset>
      </batchtest>

      <formatter type="xml" todir="testreport" outfile="logfile.xml" />
    </phpunit>

        <phpunitreport infile="logfile.xml"
                   format="frames"
                   todir="testreport"/>


  </target>
</project>
```

---

<phpunitreport> is successfully executing and creating html base report on a different system with PHP 5.4.45-0+deb7u2.

[...]

PR #650, opened 2017-01-17
Registered PHP Functions for XSL #584

siad007:
Related to #584


mrook:
Looks good! I'm thinking we may want to restrict the available php functions in `registerPHPFunctions` (possibly just to `nl2br`), what do you think?

siad007:
@mrook 👍 restricted to `nl2br`

PR #661, opened 2017-02-04
Added missing include of ComponentHelper

siad007:


mrook:
Can also be solved by `composer install` or `composer dumpautoload`, but we're probably not fully ready to get rid of the `include`/`require` statements.

PR #663, opened 2017-02-05
Fixed test execution on windows

siad007:


mrook:
Thanks! We probably need to figure out a way to automatically run the tests on Windows, travis-ci doesn't support that.

siad007:
@mrook Maybe https://www.appveyor.com would be of any help - i will check it and report if useful.

PR #665, opened 2017-02-05
Solved #654 : Auto-discover custom tasks/types through composer.

### LINKED ISSUE #654: Auto-discover custom tasks when installed through Composer
Phing should be able to auto-discover custom tasks when Phing itself is installed using Composer, and the custom tasks are part of packages installed using Composer.

siad007:
Added auto-discover function for custom types / tasks.
I have added an example task and type in composers require-dev section (`mehr-als-nix/custom-phing-task-example`) for testing purposes. After updating composer dependencies you could test the auto-discovering function with this build target:
```xml
<target name="custom-example">
	<exampletask msg="hello world" />
	<exampletype id="exampletypeid" data="foo bar" />
	<exampletask exampletyperef="exampletypeid" />
</target>
```
Fixes #654

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,3 +10,5 @@ vendor
 bin/*
 !bin/pear*
 !bin/phing*
+/custom.tasks.properties
+/custom.types.properties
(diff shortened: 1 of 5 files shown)


mrook:
Interesting! I was however thinking more along the lines of querying the Composer API for installed packages and perhaps discovering available tasks that way; possibly utilizing the `extra` field in the `composer.json` for every package that's offering custom tasks. What do you think?

PR #669, opened 2017-02-21
file_get_contents in windows return a permission error if the file is a dir. Check if file is a dir before file_get_contents

Parro:


Parro:
Sorry, my code doesn't work, I submitted the pull request by mistake

PR #672, opened 2017-02-23
Add possibility to set the phing.dir with a environment variable.

jalogut:
This will allow to execute phing projects globally. So I can do something like that on my custom phing project installed globally.

```
#!/usr/bin/env php
<?php

/**
 * Shell wrapper for Phing
 */

// turn off html errors
ini_set('html_errors', 'off');

define('BP', realpath(dirname(__DIR__)));

putenv("PHING_DIR=".BP);
putenv("PHING_HOME=" . BP . '/vendor/phing/phing');

require_once getenv("PHING_HOME") . '/bin/phing.php';
```

### DIFF
--- a/classes/phing/Phing.php
+++ b/classes/phing/Phing.php
@@ -567,6 +567,8 @@ class Phing
             // but -find then search for it
             if ($this->searchForThis !== null) {
                 $this->buildFile = $this->_findBuildFile(self::getProperty("user.dir"), $this->searchForThis);
+            } else if (getenv("PHING_DIR")) {
+                $this->buildFile = new PhingFile(getenv("PHING_DIR"), self::DEFAULT_BUILD_FILENAME);
             } else {
                 $this->buildFile = new PhingFile(self::DEFAULT_BUILD_FILENAME);
             }


jalogut:
Hello,

Any chance that this will be reviewed soon?

Thanks

mrook:
Hi @jalogut, I'm not entirely sure why you want to override phing.dir?

jalogut:
Hi @mrook, this change adds the possibility to deliver standalone applications based on PHing. That  way, we could create apps based on PHing that can be installed in other projects with composer. 
That would be similar to what can be done in Robo here:
- http://robo.li/framework/

Here an example of what I want to accomplish. This is an standalone tool created with PHing:
- https://github.com/staempfli/magento2-deployment-tool
For the moment I need to override `classes/phing/Phing.php` to add the option of overriding `phing.dir` until this PR gets merged.

Hope that is a bit clearer.

mrook:
Okay, I understand. Looking at your code, a `phing -f YOUR_PHING_DIR/build.xml` would work just as well right? Why not create a wrapper (shell) script that does just that?

jalogut:
Well, I tried that at first but it doesn't really work because then I lose the capabilities to execute phing commands as usual. So if I want to run `phing -l` from the command line, what should I do in the "wrapper shell"? Take all args, add the default `-f YOUR_PHING_DIR/build.xml` and call phing with the new args? 

I think that this would be a bit hacky. I just thought that this PR would be a nice feature to include.

mrook:
The wrapper script would essentially be `phing -f YOUR_PHING_DIR/build.xml "$@"` right?

I think the PHING_DIR override is interesting, but probably too specific to your use case to include here.

jalogut:
Hi @mrook ,

Ok, yes that seems to work. Thanks for the suggestion. I changed my bin entry point with the following shell wrapper:

```
#!/bin/bash

SOURCE="${BASH_SOURCE[0]}"
while [ -h "$SOURCE" ]; do # resolve $SOURCE until the file is no longer a symlink
  DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )"
  SOURCE="$(readlink "$SOURCE")"
  [[ $SOURCE != /* ]] && SOURCE="$DIR/$SOURCE" # if $SOURCE was a relative symlink, we need to resolve it relative to the path where the symlink file was located
done
DIR="$( cd -P "$( dirname "$SOURCE" )" && pwd )/../"

phing -f $DIR/build.xml "$@"
```

If you think, that is a better solution than adding a custom PHING_DIR with php, you can close this issue

mrook:
Thanks for the feedback @jalogut! Closing this issue for now.

PR #649, opened 2017-01-16
Fixed Support <dirset> in chmod, chown, delete, echo, copy, foreach and move task

siad007:
Fixed #559 

Additions:
- dirset documentation
- grammer fileset, dirset, ...
- foreach task new attributes
 -- inheritall
-- inheritrefs
-- trim
- foreach task new nested elements
-- path
-- reference
-- param (alias of property).
- phingcall task new attributes
-- inheritall
-- inheritrefs

PR #673, opened 2017-02-25
Fixed clone method in dirset. Related to #649

siad007:
Related to #649 :
Removed the return statements inside the `__clone` method.

### DIFF
--- a/classes/phing/types/DirSet.php
+++ b/classes/phing/types/DirSet.php
@@ -46,9 +46,9 @@ class DirSet extends AbstractFileSet
     public function __clone()
     {
         if ($this->isReference()) {
-            return new DirSet($this->getRef($this->getProject()));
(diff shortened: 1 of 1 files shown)


classes/phing/types/DirSet.php
```diff
@@ -46,9 +46,9 @@ public function __construct($dirset = null)
     public function __clone()
     {
         if ($this->isReference()) {
-            return new DirSet($this->getRef($this->getProject()));
+            new DirSet($this->getRef($this->getProject()));
         } else {
-            return new DirSet($this);
+            new DirSet($this);
```
mrook: Hm does this actually do anything? I'm not sure we even need to keep the `__clone` methods to be honest, what do you think?

mrook:
Closing as the `__clone` methods were removed in bd4f6bb.

PR #677, opened 2017-02-25
Made Reference aware of Project Related to #649

siad007:
Related to #649 :
Made Reference aware of Project and give optional the ability to overwrite project in `getReferencedObject`


siad007:
Obsolete

Issue #684, opened 2017-03-03
Relative Symlinks


dwkitchen:
I wanted to create Symlinks that would work on my local machine and in a VM, so I was happy to see the addition of relative symlinks.

However, I have found that the solution added in #567 does not work. The target used is relative to the root of the project folder, not relative the to the link being created.

If in this example ther is a file index.php:
```
<symlink link="dest" relative="true">
  <fileset dir="src">
    <include name="*" />
  </fileset>
</symlink>
```

The link the is created is `dest/index.php -> src/index.php` but this is wrong for a symlink, it should be `dest/index.php -> ../src/index.php`.

For relative links to work the depth of the *link* from the project base needs to be calculated and the correct number of `../` added to the front of *target*. I have created an initial WIP, but this may need more work to refine possible use cases.

dwkitchen:
As I investigate options I see that in the example in the SymlinkTask.php file is

```
<code>
    <symlink link="${project.basedir}/library">
        <fileset dir="/my/shared/library">
            <include name="*" />
        </fileset>
    </symlink>
</code>
```

The recommendation to include `${project.basedir}` at the start of the *link* path makes this more complicated also the use of an absolute path for the source means that this would not work in my example of the symlinks working inside a VM.

I am wondering if there is already a function in the Phing dependencies that will calculate a relative path?

Is it worth adding symfony/filesystem for https://github.com/symfony/Filesystem/blob/00e0462cfe48c7fdd142d0867e827cb0d17e4e23/Filesystem.php#L327 ?

mrook:
#685 also tries to solve this.

mrook:
#695 also tries to solve this.

mrook:
Closed by merging #695.

PR #685, opened 2017-03-03
#684 Relative Symlinks

dwkitchen:
Based on Symfony/Filesystem->makePathRelative()


classes/phing/tasks/ext/SymlinkTask.php
```diff
@@ -236,7 +236,36 @@ protected function getMap()
             }
 
             if ($this->isRelative()) {
-                $fromDir = $fs->getDir($this->getProject())->getPath();
+                $endPath = $fs->getDir($this->getProject())->getAbsolutePath();
+                $startPath = $this->getLink();
```
dwkitchen: This only works if the link path given is absolute, how to get absolute path from `$this->getLink()?

siad007: You can get the absolute path for `$this->getLink()` in combination with `PhingFile` like:
```
$startPath = (new PhingFile($this->getLink()))->getAbsolutePath();
```

dwkitchen: @siad007 Thanks for that updated and working with a relative path to the project root is given in the command.

mrook:
Closed in favor of #695 - thanks for your contribution!

PR #689, opened 2017-03-05
Added optional nested zipfileset for ZipTask.

siad007:
Added optional use of a real zipfileset nested element leaving the FileSet support intact.


classes/phing/tasks/ext/ZipTask.php
```diff
@@ -172,7 +179,7 @@ public function main()
 
                 if (empty($this->filesets)) {
                     // add the main fileset to the list of filesets to process.
-                    $mainFileSet = new ZipFileSet($this->fileset);
+                    $mainFileSet = $this->fileset;
```
mrook: `$this->fileset` is no longer set in `createFileSet`

siad007:
@mrook fixed

PR #692, opened 2017-03-05
Added parent exception argument to setError for selectors.

siad007:


siad007:
@mrook I don't think that this is a bc break. All setters in phing which goes through the introspection have exactly 1 parameter except for the `setError` method call. So everything should be fine here :)

Issue #558, opened 2016-10-09
ComposerTask when composer is installed in the system (Trac #1008)


phing-issues-importer:
Is tricky to run the ComposerTask when composer is installed in the system instead of in project directory.
I propose to enhance the task in order to modify the command that is passed to `passtruh`.

Maybe using by default the system's composer installation, identify it through `which composer` and override the behaviour if composer.phar is present in project dir.

Migrated from https://www.phing.info/trac/ticket/1008

``` json
{
    "status": "new", 
    "changetime": "2015-11-28T19:10:43", 
    "description": "Is tricky to run the ComposerTask when composer is installed in the system instead of in project directory.\nI propose to enhance the task in order to modify the command that is passed to `passtruh`.\n\nMaybe using by default the system's composer installation, identify it through `which composer` and override the behaviour if composer.phar is present in project dir.", 
    "reporter": "giacomorama@gmail.com", 
    "cc": "", 
    "resolution": "", 
    "_ts": "1448737843092041", 
    "component": "phing-tasks-ext", 
    "summary": "ComposerTask when composer is installed in the system", 
    "priority": "minor", 
    "keywords": "composer", 
    "version": "2.5.0", 
    "time": "2013-04-22T12:19:05", 
    "milestone": "Backlog", 
    "owner": "mrook", 
    "type": "enhancement"
}
```

phing-issues-importer:
Trac update at 20130507T08:25:42: mrook commented:

> Unfortunately 'which composer' is not going to work on a Windows machine.

phing-issues-importer:
Trac update at 20140606T09:35:46: anonymous commented:

> On windows there is where.exe...

phing-issues-importer:
Trac update at 20140606T09:38:29: anonymous commented:

> Replying to [comment:2 anonymous]:
> > On windows there is where.exe...
> Forgettig to say that just not work on WinzXP, where there are other solutions.
> Just to say that is possible to implement.
>

phing-issues-importer:
Trac update at 20140919T10:37:38: anonymous commented:

> I'm also looking for a solution to use a global installed composer. I think most developers are using composer like this and therefore phing should support that. Of course providing the full path in the build.xml is not an option.
> 
> The simplest way to check, if there's a global composer installation would be to exec "composer about" or something like that and check the return code and maybe the output. 
> 
> Using "which" or "where.exe" would be more of a challenge to make it work on all platforms. 
> 
[...]

PR #701, opened 2017-03-24
Dynamic path for composer task

verbruggenalex:
It would be a nice feature to have the phing composer task check if a composer installation is available to execute the command with if no composer executable path is set.


mrook:
Related to #558.

verbruggenalex:
Would it be a good idea to implement a try catch on the command with the provided composer path and then try finding a path if that fails? That would cover the user entering a wrong path. But then the user would not be aware of that mistake.

verbruggenalex:
I moved the path check to the getComposer function and cleaned up the class a bit.

verbruggenalex:
Mind giving me a new review @mrook ?

mrook:
Can you squash the commits? The PR is a little noisy this way :)

Anyway:

* I see a lot of unrelated changes (docblock modifications etc.)
* Instead of checking against `php_uname` you can perhaps reuse the `host.fstype` property

PR #336, opened 2014-12-20
Added property file generation for VersionTask

siad007:
Added property file generation and made file name and property optional.

PR #713, opened 2017-05-03
Fix for issue #771 - VersionTask output file contains trailing new line character

chrisjohnson00:
updated VersionTaskTest to reproduce issue #771
update to VersionTask to no longer append the property line separator to the resulting version file, test passes


siad007:
Already implemented in #336 . So i will close this.

PR #714, opened 2017-05-03
Add "which" method to FileSystem class, for solving Issue 710.

kenguest:
This could also be used in the rST task, instead of requiring PEAR's System class.
@cweiske - what do you think? I value your opinion as you are both the originator of the rST task, a phing maintainer and a stalwart of the PEAR project :-)


cweiske:
Sure, do it. I'm not hanging onto System.php in phing.

kenguest:
Done - and all squashed back into one commit, for whoever's reviewing pleasure :-)

PR #726, opened 2017-06-15
Fix class name of XML Logger issue #724

arruor:
Fixing class name for XMLFormatter class.


arruor:
Closed due to wrong PHPLoc version in composer.json.

Issue #725, opened 2017-06-15
PHPLoc Task: Wrong class name in CSV Formatter


arruor:
composer.json:
```json
"require-dev": {
        "phploc/phploc": "^4.0",
        "phing/phing": "^2.16"
    },
```

task definition:
```xml
<target name="phploc" unless="phploc.done"
            description="Measure project size using PHPLOC and print human readable output">
        <phploc countTests="true">
            <fileset refid="phpFiles"/>
            <formatter type="cli" outfile="build/html/phploc.txt"/>
            <formatter type="cli" outfile="build/logs/phploc.txt"/>
            <formatter type="xml" outfile="build/logs/phploc.xml"/>
            <formatter type="csv" outfile="build/logs/phploc.csv"/>
        </phploc>
        <property name="phploc.done" value="true"/>
    </target>
```
Stack trace
```php
[phploc] Writing report to: ./build/logs/phploc.xml
   [phploc] Writing report to: ./build/logs/phploc.csv
PHP Fatal error:  Uncaught Error: Class '\SebastianBergmann\PHPLOC\Log\CSV\Single' not found in /path/to/project/vendor/phing/phing/classes/phing/tasks/ext/phploc/PHPLocCSVFormatter.php:34
Stack trace:
#0 /path/to/project/vendor/phing/phing/classes/phing/tasks/ext/phploc/PHPLocTask.php(310): PHPLocCSVFormatter->printResult(Array, true)
#1 /path/to/project/vendor/phing/phing/classes/phing/tasks/ext/phploc/PHPLocTask.php(219): PHPLocTask->runPhpLocCheck()
#2 /path/to/project/vendor/phing/phing/classes/phing/UnknownElement.php(100): PHPLocTask->main()
#3 /path/to/project/vendor/phing/phing/classes/phing/Task.php(283): UnknownElement->main()
#4 /path/to/project/vendor/phing/phing/classes/phing/Target.php(336): Task->perform()
#5 /path/to/project/vendor/phing/phing/classes/phing/Target.php(366): Target->main()
#6 /path/to/project/vendor/phing/phing/classes/phing/Project.php( in /path/to/project/vendor/phing/phing/classes/phing/tasks/ext/phploc/PHPLocCSVFormatter.php on line 34

[...]

PR #727, opened 2017-06-15
Fix class name of CSV Logger issue #725

arruor:
Fixing class name of CSV logger in CSV Formatter class.


arruor:
Closed due to wrong PHPLoc version in composer.json.

PR #740, opened 2017-07-10
Default exclude task

siad007:
Added DefaultExclude task.
Made an updated PR for this with newly added tests and grammer changes.

Closes #371


mrook:
@siad007 can you resolve the conflicts here?

PR #747, opened 2017-07-31
Demonstrate StringHelper::booleanValue in the custom MyEchoTask code.

kenguest:
...rather than having people code their own "yes|y|no|n|true|t|1|0" to boolean conversion code...

### DIFF
--- a/docs/docbook5/en/source/chapters/extending.xml
+++ b/docs/docbook5/en/source/chapters/extending.xml
@@ -368,12 +368,20 @@ class MyEchoTask extends Task {
      * The message passed in the buildfile.
      */
     private $message = null;
+    /**
+     * Whether to reverse the message, for fun?
+     */
+    private $reverse = false;
 
     /**
      * The setter for the attribute "message"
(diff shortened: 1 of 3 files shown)


docs/docbook5/en/source/chapters/extending.xml
```diff
@@ -376,6 +380,10 @@ class MyEchoTask extends Task {
         $this->message = $str;
     }
 
+    public function setReverse($str) {
+        $this->reverse = StringHelper::booleanValue($str);
```
siad007: It is not necessary to use `StringHelper::booleanValue()` here as it happens in the
[`IntrospectionHelper`](https://github.com/phingofficial/phing/blob/ac3cc900a34fd0a2647850f367268687135b030b/classes/phing/IntrospectionHelper.php#L326) as well.

kenguest: I'd be inclined to agree but 'y' or 'n' values support...

PR #751, opened 2017-08-06
Additional work done to NotifySend task.

kenguest:
* Amend for PSR2 Coding Style.
* Use StringHelper::booleanValue instead of blindly casting to boolean value (re Silent attribute).
* Use new FileSystem which method to determine if executable is available.
* Extend debug logging messages.

### DIFF
--- a/classes/phing/tasks/ext/NotifySendTask.php
+++ b/classes/phing/tasks/ext/NotifySendTask.php
@@ -39,25 +39,25 @@ class NotifySendTask extends Task
     public function setIcon($icon)
     {
         switch ($icon) {
-        case 'info':
-        case 'error':
-        case 'warning':
-            $this->icon = $icon;
-            break;
-        default:
-            if (file_exists($icon) && is_file($icon)) {
+            case 'info':
+            case 'error':
+            case 'warning':
                 $this->icon = $icon;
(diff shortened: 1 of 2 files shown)


classes/phing/tasks/ext/NotifySendTask.php
```diff
@@ -74,13 +74,13 @@ public function getIcon()
...
      *
-     * @param bool $silent Don't execute notifysend? True not to.
+     * @param string $silent Don't execute notifysend? Truthy value.
      *
      * @return void
      */
     public function setSilent($silent)
     {
-        $this->silent = (bool) $silent;
+        $this->silent = StringHelper::booleanValue($silent);
```
siad007: It is not necessary to use `StringHelper::booleanValue()` here as it happens in the
[`IntrospectionHelper`](https://github.com/phingofficial/phing/blob/ac3cc900a34fd0a2647850f367268687135b030b/classes/phing/IntrospectionHelper.php#L326) as well. Just dropping the bool cast would be enough.

kenguest: I'd be inclined to agree if I didn't want 'y' or 'n' values supported...

PR #753, opened 2017-08-10
Add Get tag to inifile task; set [section]key value to an property.

kenguest:
For instances where you need to set property to a value present in an ini file that has multiple sections - not just a simple phing/ant property file. Using along the lines of:

    <?xml version="1.0"?>
    <project name="foo" default="main">
    <target name="main">
        <inifile source="get.ini" verbose="yes">
        <get section="test" property="foo" outputproperty="qux"/>
        </inifile>
    <echo msg="${qux}"/>
        <inifile source="get.ini" verbose="yes">
            <get section="test2" property="foo" outputproperty="qux" default="omg"/>
        </inifile>
    <echo msg="${qux}"/>
    </target>
    </project>

Includes some small fixes to docbook source that was preventing the updated docs w IniFile 'Get' documentation from building...


mrook:
@kenguest can you resolve the conflicts?

PR #759, opened 2017-08-20
Added location setting to all task defined.

siad007:
Every time a task is proceeded the location would be set.


mrook:
@siad007 can you fix the conflicts?

siad007:
Was added in #741

PR #761, opened 2017-08-21
Fix: php.ini variable evaluation and "Notice: A non well formed numeric value encountered"

matyasb:
The php.ini can contain values like: '1K' (as 1 kilobyte) instead of '1024' or '1M' as (1 megabyte) instead of '1048576'.
For example:
memory_limit = 32M
post_max_size = 32M

Phing is buggy as '1M' is handled as same as '1K'.

Additionally $val = '2k' * 1024 cause a "Notice: A non well formed numeric value encountered".
Before PHP 7 it was ignored and handled as 2 * 1024.

### DIFF
--- a/classes/phing/Phing.php
+++ b/classes/phing/Phing.php
@@ -1580,21 +1580,23 @@ class Phing
      * @param string $val
      * @return int
      */
-    private static function convertShorthand($val)
+    public static function convertShorthand($val)
     {
         $val = trim($val);
         $last = strtolower($val[strlen($val) - 1]);
 
-        $val = (int) $val;
+        if (!is_numeric($last)) {
+            $val = (int) substr($val, 0, strlen($val) - 1);
 
-        switch ($last) {
-            // The 'G' modifier is available since PHP 5.1.0
-            case 'g':
-                $val *= 1024;
-            case 'm':
-                $val *= 1024;
-            case 'k':
-                $val *= 1024;
+            switch ($last) {
+                // The 'G' modifier is available since PHP 5.1.0
+                case 'g':
+                    $val *= 1024;
+                case 'm':
+                    $val *= 1024;
+                case 'k':
+                    $val *= 1024;
+            }
         }
 
         return $val;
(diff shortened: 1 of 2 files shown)


classes/phing/Phing.php
```diff
@@ -1584,15 +1584,16 @@ private static function convertShorthand($val)
...
-        $val = (int) $val;
+        if (!is_numeric($last)) {
+            $val = substr($val, 0, strlen($val) - 1);
+        }
 
         switch ($last) {
             // The 'G' modifier is available since PHP 5.1.0
             case 'g':
-                $val *= 1024;
+                $val *= 1024 * 1024 * 1024;
```
siad007: Would you mind adding a test for this switch? Without break statements it seems wrong for me here? What do you think?

matyasb:
You are right about this is not correct because the missing break mislead me. I focused only the notice.

kenguest:
This needs to be fixed before being merged - either remove the changes to the $val *=  1024... lines or add use of the break keyword to the switch statement. I would favour the first remedy.

Otherwise a 'g' suffix will result in $val being multiplied by 1024 * 1024 * 1024 * 1024 * 1024 and a 'm' suffix will result in $val being multiplied by 1024 * 1024 * 1024.

This checklist might help:
- [ ] Fix switch statement.
- [ ] Add a test for the 'g' suffix.
- [ ] Add a test for the 'm' suffix.
- [ ] Add a test for the 'k' suffix.
- [ ] Add a test for a nonsense suffix, maybe 'j'.
- [ ] Add a test for no suffix; -1 and at least one other.

Thanks, @matyasb, for wanting to improve the quality of phing - you just need to be a little more careful though ;-)

matyasb:
I choosed "return" statement instead of "break" like parseHumanSizes in tasks\system\condition\HasFreeSpaceCondition.php

I think this way is more understandable.

kenguest:
That looks great now - I see you've added tests and tightened up the code - thanks for your hard work @matyasb!

matyasb:
Thanks for your help.

siad007:
@matyasb please squash your commits for merging.

mrook:
Thanks for the contribution @matyasb !

matyasb:
Sorry I don't want to revert!

mrook:
No problem :-)

PR #769, opened 2017-09-01
SCA with Php Inspections (EA Extended)

kalessil:
Performance tweaks and minor simplifications.

PS: `$xml = simplexml_load_string(file_get_contents($filePath));` is a bug-fix for https://bugs.php.net/bug.php?id=62577

### DIFF
--- a/classes/phing/Phing.php
+++ b/classes/phing/Phing.php
@@ -1317,8 +1317,7 @@ class Phing
             self::$importPaths = self::explodeIncludePath();
         }
 
-        $path = str_replace('\\', DIRECTORY_SEPARATOR, $path);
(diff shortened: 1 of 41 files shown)


kalessil:
Seems 0413f7b break build, will revert it later.

kenguest:
Using get_class() with no parameters would break against/with PHP7.2 anyway...

There's a maxim of "if it ain't broke, don't fix it." - something to consider! :-)

kalessil:
Closing, re-created the PR in #770

kalessil:
@kenguest, this perfectly applies to legacy code bases =) 
I think it worth contributing something back to this great project =)

PR #785, opened 2017-10-01
[WIP] Small improvement on comparing files.

siad007:
Reduces the call to file_get_contents


siad007:
@kenguest changed the order thx

siad007:
@kenguest there seems to be a trim issue - better give it a new attribute for text based or binary based comparison here to avoid that kind of misbehavior. i will have a closer look on that here ...

Issue #796, opened 2017-10-20
Init feature


jawira:
It would be nice to have an "initialize" option for Phing.

I systematically look for this feature everytime I discover a new package.

This kind of feature is usual nowadays, to name a few:

  * Behat: `$ behat --init`
  * Composer: `$ php composer.phar init`
  * Phinx: `$ phinx init .`

So for Phing I'm wanting something like this:

```sh
$ vendor/bin/phing -init
```

```sh
$ vendor/bin/phing -i
```

Ideally this command would create two files at the root of the project: `build.xml` and `build.properties` (these are the default naming conventions).

The content of `build.xml`:

```xml
<?xml version="1.0" encoding="UTF-8" ?>

<project name="" description="" default="">
    
    <target name="" description="">
        
    </target>
    
</project>
```

The content of `build.properties`:

```ini
[section]
name=value
```

I have already worked on this, and I'm really expecting for your feedback.

siad007:
Thanks for your PR!
Some small issues:

- do we really need a property file generated?
- if so, there is no support for sections by a property file but for ini files

I would suggest dropping the generation of the property file.

siad007:
What do you think about a optional usage of a filename?
```
$ vendor/bin/phing -init "phing.xml"
```
It would also be nice to support
- a single file (phing.xml)
- with a relative path (test/phing.xml)
- or an absolute path (/path/to/phing.xml)
What do you think.

Jawira:
Thanks a lot for your quick feedback!

About **generation of property file**: I added this because it's what I always use in all my projects, and I consider it's a good practice to keep all properties together in a same file, however, I agree that not all people need or want a property file.

Regarding sections, I know they are not supported by properties and are omitted, but I always use sections in my properties as categories:
```ini
[database]
db.user=user

[ftp]
ftp.server=

[tests]
dir.output=
```
But once again, I agree this couldn't be the case for all Phing's users. Therefore I will drop the generation of property file as you suggest, mainly for the sake of simplicity.

Concerning the **optional usage of a filename**, I'm not very keen about this. The main reason is that I want to keep the `-init` option really simple and straight forward. I always call my buildfile `build.xml` because this way I don't have to use the `-f` option and the command is shorter in terminal. Even if I'm not really against this idea I prefer keeping things simple.

Jawira:
I updated my PR, I have taken into account your suggestions.

mrook:
Closed because #797 was merged.

PR #797, opened 2017-10-20
Feature/init option

jawira:
Please see: phingofficial/phing#796


classes/phing/Phing.php
```diff
@@ -70,6 +70,7 @@ class Phing
 
     /** The default build file name */
     const DEFAULT_BUILD_FILENAME = "build.xml";
+    const DEFAULT_PROPERTIES_FILENAME = "build.properties";
```
siad007: Do we really need a build.properties file?

classes/phing/Phing.php
```diff
@@ -1048,6 +1055,56 @@ public static function printVersion()
...
+    }
+
+    /**
+     * Creates generic Property file
+     *
+     * @param string $dir
+     */
+    protected static function initPropertyFile($dir)
+    {
+        $content = '[section]' . PHP_EOL;
```
siad007: Sections are not supported by property files but by ini files.

jawira:
Changes have been made :)

classes/phing/Phing.php
```diff
@@ -1048,6 +1055,38 @@ public static function printVersion()
     }
 
     /**
+     * Creates generic buildfile
+     */
+    public static function init()
+    {
+        $homeDir = self::getProperty(self::PHING_HOME);
```
siad007: I think we could take the app start directory here `self::getProperty('application.startdir')`

jawira:
Regarding the default directory, my intention was to create de sample build.xml in the current directory as you mention. I will correct this.

About overriding default location and/or file name, I wanted to keep it simple but I agree it's a nice feature to have. I will start working on this feature.

jawira:
Hello guys, now -init option accepts an optional path, it can be a directory or a file name.
Please let me know if you have any question. I'm waiting for your feedback too :)

mrook:
Looks good! Except for two last things. One very minor, it looks like some accidental white space changes slipped into the PR. Lastly, would it be nice to throw an exception when the (specified) build file already exists? Rather than silently stopping.

jawira:
Hello, thanks for your comments. I have done the requested changes :)

PR #798, opened 2017-10-21
Fixed generation of html reportfiles.

siad007:
if more than one testsuite is in junit.xml

This is more likely a fix to render multiple testsuites correctly.
There might be a better support for namespaces and packages but it is not part of this PR.


joel-depiltech:
So good to hear that !

Issue #799, opened 2017-10-23
Can't install dev-master version using Composer


MaXal:
I have the following `composer.json`:
```
{
  "require-dev": {
    "phing/phing" : "dev-master"
  }
}
```

Running `composer install` I get:
Your requirements could not be resolved to an installable set of packages.

```
  Problem 1
    - Installation request for phing/phing dev-master -> satisfiable by phing/phing[dev-master].
    - phing/phing dev-master requires phpdocumentor/parallel master@dev -> no matching package found.
```

I believe the reason is the following commit: 97532911e70c0a87ddc452648a40dced861b2b3b

siad007:
hi @MaXal can you please try adding following to your composer.json - try again and post the result here again please?
```
    "repositories": [
        {
            "type":"package",
            "package": {
                "name": "phpDocumentor/Parallel",
                "version":"master",
                "source": {
                    "url": "https://github.com/phpDocumentor/Parallel.git",
                    "type": "git",
                    "reference":"master"
                }
            }
        }
    ],
    "require-dev": {
        "phpDocumentor/Parallel": "master@dev",
        "phing/phing" : "dev-master"

```

mwierzbi:
nope. after added repository there is an error when composer tries to generate autoload:

```
Generating autoload files

                                                                                                                  
  [RuntimeException]                                                                                              
  Could not scan for classes inside "/home/test/vendor/phing/phing/vendor/phpDocumentor/Paralle  
  l" which does not appear to be a file nor a folder
```

I think this line is a problem:
https://github.com/phingofficial/phing/blob/master/composer.json#L68

MaXal:
@siad007 I get the same error as @mwierzbi and to be honest I believe there should be no need to add additional dependency to the client composer.json anyway since it defeat the purpose of Composer.

siad007:
@mwierzbi @MaXal I will do a PR on the `phpDocumentor/Parallel` repo for composer support.

mrook:
Should we revert the change in phing so that dev-master works again?

siad007:
Let us revert the change until phpDocumentor/Parallel will served by composer. Thx @mrook

MaXal:
@siad007 Any news on the issue?

mrook:
@MaXal #800 was just merged, can you test & verify that that solves it?

MaXal:
@mrook Having the following `composer.json`:
```
{
  "name": "vendor_name/package_name",
  "description": "description_text",
[...]

PR #800, opened 2017-10-27
Define autoload for phpDocumentor/Parallel in that repository.

bangpound:
The repository package definition for the `phpDocumentor/Parallel` dependency can and should
include its own autoload classmap rather than configuring the autoloader of this project to look
in its own vendor directory. This change will allow phing to be a dependency of other projects.


siad007:
Related to #799

PR #812, opened 2017-12-02
2.16.1 proposal

adam-rocska:
# Bugfix for PHPUnitTask class

**Expected behavior** : PHPUnit runs as expected
**Actual behavior** : PHPUnit fails with `Argument 1 passed to File_Iterator_Factory::getPathsAfterResolvingWildcards() must be of the type array, boolean given`
**How to reproduce** : Use the PHPUnit task to run tests, and expect coverage reporting without a blacklist
**Root Cause** : realpath returned false for in-phar dir


classes/phing/tasks/ext/phpunit/PHPUnitTask.php
```diff
@@ -482,7 +482,7 @@ protected function execute($suite)
             /**
              * Add some defaults to the PHPUnit filter
              */
-            $pwd = __DIR__;
+            $pwd = getcwd();
             $path = realpath($pwd . '/../../../');
```
siad007: I think the main problem is the use of `realpath()` function inside a phar archive.
Could you please try with:
```php
$pwd = __DIR__;
$path = dirname($pwd, 3) . '/';
```

PR #855, opened 2018-01-20
Fixed wrong init value

siad007:


classes/phing/filters/ConcatFilter.php
```diff
@@ -87,7 +87,7 @@ public function __construct(Reader $in = null)
      *                     during reading
      * @throws BuildException
      */
-    public function read($len = 0)
+    public function read($len = null)
```
Yousha: It's unused... isn't?

PR #857, opened 2018-01-20
[WIP] Some cleanup and preparation to make Phing class overwritable.

siad007:
Added PhingMain interface and modifications to allow overwriting the Phing class in the future.

### DIFF
--- a/classes/phing/Phing.php
+++ b/classes/phing/Phing.php
@@ -18,7 +18,6 @@
  */
 
 use SebastianBergmann\Version;
-use Symfony\Component\Console\Input\StringInput;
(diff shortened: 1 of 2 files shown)


mrook:
Can you explain the use case?

siad007:
@mrook the main use case here would be the implementation of a [Launcher](https://github.com/apache/ant/blob/master/src/main/org/apache/tools/ant/launch/Launcher.java) which takes care of loading a main class. So it would be possible to have a custom main class and/or a custom launcher.

mrook:
Hm that feels a little premature? Maybe wait until we have a concrete use case for that?

PR #875, opened 2018-02-27
Add delta to foreachtask in loop

sgurlt:
This allows the user to use the delta of the foreach loop within the called phing task. The default property name for the delta is "delta", but could be overridden with in the foreach call.

### DIFF
--- a/classes/phing/tasks/system/ForeachTask.php
+++ b/classes/phing/tasks/system/ForeachTask.php
@@ -63,6 +63,9 @@ class ForeachTask extends Task
     /** Delimiter that separates items in $list */
     private $delimiter = ',';
 
+    /** Delimiter that separates items in $list */
+    private $delta = 'delta';
+
     /**
      * PhingCallTask that will be invoked w/ calleeTarget.
      * @var PhingCallTask
@@ -139,7 +142,7 @@ class ForeachTask extends Task
             $arr = explode($this->delimiter, $this->list);
             $total_entries = 0;
 
-            foreach ($arr as $value) {
+            foreach ($arr as $delta => $value) {
                 $value = trim($value);
                 $premapped = '';
                 if ($mapper !== null) {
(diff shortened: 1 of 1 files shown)


classes/phing/tasks/system/ForeachTask.php
```diff
@@ -63,6 +63,9 @@ class ForeachTask extends Task
     /** Delimiter that separates items in $list */
     private $delimiter = ',';
 
+    /** Delimiter that separates items in $list */
+    private $delta = 'delta';
```
mrook: There's no setter for `$delta` (which will be needed to override the default value).

mrook: And while we're at it, please document the new property in `docs/guide/source/en` as well.

mrook:
Thanks for your contribution!

sgurlt:
Good call on the setter for delta, I just forgot to add it 👍 
According to the docs, I just had a look on the structure you send me, but unfortunately, there is not guide folder? https://github.com/phingofficial/phing/tree/oldstable/docs

mrook:
Ah I missed that you based the PR on `oldstable`. Can you rebase it on `master`? The correct location for the docs would be `https://github.com/phingofficial/phing/blob/master/docs/guide/en/source/appendixes/coretasks.xml`

sgurlt:
I will first have to check of the update to the master branch does not break my tasks in general. If that works without any issues, I will create a new pull requests related to the master branch.

mrook:
There hasn't been a 3.x release yet, which is why there's no tag :-) As for the 2.x series, I don't plan to release any more updates other than important / high impact (bug) fixes.

sgurlt:
Damn, as expected is one of the packages I am using to integrate phing with drupal drush: https://www.drupal.org/project/phingdrushtask depending the 2.x Release of phing.

Since it will take a moment longer to recheck the implementation, I am going to move that to the weekend. I will create a patch for the Drupal integration and will give it a try with your the latest master release of phing.
I will keep you updated.

PR #890, opened 2018-03-26
Fix PHP CodeSniffer cache write to directory

ruifil:
Fix error of PHP_CodeSniffer_Reports_PhingRemoveFromCache try to write to content to directory path :
[PHP Error] file_put_contents(/data/): failed to open stream: Is a directory [line 191 of /data/library/vendor/squizlabs/php_codesniffer/CodeSniffer/Reporting.php]


siad007:
@ruifil Thanks for your contribution.

PR #922, opened 2018-05-22
Small additions to the os condition

siad007:


siad007:
But this is a addition to the OsCondition so why should this condition not be able to determine a netware system?

kenguest:
A quick search for "PHP  on NetWare" shows that "PHP for NetWare® is based on open source PHP version 4.2.3 and PHP version 5.0.5.".

Unless that is grossly inaccurate and PHP  7 can run on that discontinued operating system, I think you are adding code that will never execute.

mrook:
I have the same comments about "dos". So, let's stick to detecting linux/mac/windows?

siad007:
Hmm there seems to be a problem with a test from https://github.com/phingofficial/phing/pull/907 on https://travis-ci.org/phingofficial/phing/jobs/387075667 - strange that this test was not failing before?! Any hints?

mrook:
@siad007 it did actually fail before, but the failure was ignored for a while because of the codecov upload. See https://github.com/phingofficial/phing/commit/8aea6753bd903e81acbfedf2cb7beeafaff556f2#diff-b4553af39acae935d505c76e139d4aa3.

siad007:
@mrook it becomes stranger after executing the failing test on my environment (windows) with no failure.

mrook:
@siad007 yeah I've noticed the test runs fine when run on its own, but fails when the entire test suite is run. Side-effect of another test? Cleanup gone wrong?

PR #926, opened 2018-05-23
Added nested params to PhpEvalTask

siad007:
PhpEvalTask with nested params.
Example usage:
```xml
        <target name="help">
		<php class="phing.Phing" function="start">
			<param>
				<param value="-projecthelp"/>
				<param value="-buildfile"/>
				<param value="${phing.file}"/>
			</param>
		</php>
	</target>
```


mrook:
What does this solve / what is it intended for?

siad007:
We are able to use nested params as array. So we would be able to call
methods which requires array arguments.

Michiel Rook <notifications@github.com> schrieb am Do. 24. Mai 2018 um
14:43:

PR #936, opened 2018-06-24
Expanding a Property Reference with ${toString:}

siad007:
Instead of a property value we would be able to invoke the `__toString` method of a reference like:

```
    <fileset id="sourcefiles" dir="src" includes="**/*.php"/>
    <echo> sourcefiles = ${toString:sourcefiles} </echo>
```

### DIFF
--- a/classes/phing/PropertyHelper.php
+++ b/classes/phing/PropertyHelper.php
@@ -146,6 +146,12 @@ class PropertyHelper
             }
         }
 
+        if (self::$project !== null && StringHelper::startsWith('toString:', $name)) {
+            $name = StringHelper::substring($name, strlen('toString:'));
+            $v = self::$project->getReference($name);
+            return ($v === null) ? null : (string) $v;
+        }
+
         return null;
     }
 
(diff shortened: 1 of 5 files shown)


classes/phing/PropertyHelper.php
```diff
@@ -194,7 +200,17 @@ public function replaceProperties($ns, $value, $keys)
...
-                    if (!array_key_exists($propertyName, $keys)) {
+
+                    $replacement = null;
+                    if (array_key_exists($propertyName, $keys)) {
+                        $replacement = $keys[$propertyName];
+                    }
+
+                    if ($replacement === null) {
+                        $replacement = $this->getProperty(null, $propertyName);
+                    }
```
wolzogen: Remove this, please.

classes/phing/PropertyHelper.php
```diff
@@ -194,7 +200,17 @@ public function replaceProperties($ns, $value, $keys)
...
+
+                    if ($replacement === null) {
                         self::$project->log(
```
wolzogen: and paste there:
```php
$replacement = $this->getProperty(null, $propertyName);
```

classes/phing/PropertyHelper.php
```diff
@@ -377,10 +383,15 @@ public function setInheritedProperty($ns, $name, $value)
      */
     public function getProperty($ns, $name)
     {
-        if (!isset($this->properties[$name])) {
+        if ($name === null) {
             return null;
         }
-        $found = $this->properties[$name];
+        $o = $this->getPropertyHook($ns, $name, false);
+        if ($o !== null) {
```
wolzogen:
```php
if ($o = $this->getPropertyHook($ns, $name, false)) {
   return $o;
}
```

classes/phing/PropertyHelper.php
```diff
@@ -409,14 +415,14 @@ public function getProperty($ns, $name)
      */
     public function getUserProperty($ns, $name)
     {
-        if ($name === null || !isset($this->userProperties[$name])) {
+        if ($name === null) {
             return null;
         }
         $o = $this->getPropertyHook($ns, $name, true);
         if ($o !== null) {
```
wolzogen:
```php
if ($o = $this->getPropertyHook($ns, $name, false)) {
   return $o;
}
```

PR #937, opened 2018-06-30
Fixed DefaultExcludes by removing an old hhvm fix

siad007:
As mentioned on our slack channel by @jawira there is a problem extending the default excludes list. @jawira would you mind to verify this?


jawira:
Cool, it works well for me👍

Issue #944, opened 2018-07-09
Archive task on master branch throws an error


dmromanov:
Archive task on master branch throws an error.

Phing 2bd038062732194173e200f42c076070beda61be

Build.xml
```
    <target name="create-archive" depends="build">
        <tar destfile="build/archive.tar" compression="none" includeemptydirs="true" basedir="build/build"/>
        <echo>Done.</echo>
    </target>
```

Output
```
Argument 1 passed to AbstractFileSet::getDirectoryScanner() must be an instance of Project, null given, called in /home/dromanov/Документы/Работа/sso-module/vendor/phing/phing/classes/phing/tasks/ext/TarFileSet.php on line 38
```

PR #945, opened 2018-07-09
Fix for archive task

dmromanov:
See #944


siad007:
@dmromanov Thanks for this contribution.

PR #963, opened 2018-09-29
Added index to foreach task

siad007:
Superseeds and closes #875


mrook:
Should we call it `index`, rather than `delta`?

PR #967, opened 2018-10-07
Add verbosity to VersionTask

jawira:
Using VersionTask is not so obvious at the beginning, so I added some log messages. I also did a small code refactoring.


jawira:
Yep, I fixed the bug :)

PR #972, opened 2018-10-17
Avoid calling get_class on null in UnknownElement.

elyobo:
Explicitly passing NULL to get_class triggers E_WARNING as of PHP 7.2[1].

[1] https://secure.php.net/get_class

### DIFF
--- a/classes/phing/UnknownElement.php
+++ b/classes/phing/UnknownElement.php
@@ -117,7 +117,7 @@ class UnknownElement extends Task
             $parent = $parent->getProxy();
         }
 
-        $parentClass = get_class($parent);
+        $parentClass = $parent === null ? get_class() : get_class($parent);
         $ih = IntrospectionHelper::getHelper($parentClass);
 
         for ($i = 0, $childrenCount = count($this->children); $i < $childrenCount; $i++) {


elyobo:
@siad007 I'd be happy to, but you'll have to be more clear - as far as I can see it's already compliant with the [requirements for ternaries](https://github.com/php-fig-rectified/fig-rectified-standards/blob/master/PSR-2-R-coding-style-guide-additions.md#ternary-operator), was there something else you were referring to?

siad007:
@elyobo https://github.com/php-fig/fig-standards/blob/master/accepted/PSR-2-coding-style-guide.md#25-keywords-and-truefalsenull

elyobo:
Thanks, fixed and squashed.

siad007:
@elyobo 

> If you're inclined to merge this, can we also merge it back as a fix on to the 2.x series?

it does not make sence for the 2.x branch as well exept @mrook is willing to make it happen. 
For php > 7 the actual 3.x is the way to go in my opinion.

thx for contributing.

elyobo:
Fair enough; I haven't looked at whatever changes are upcoming in 3.x, but
given the major version change I was assuming backwards incompatibility.
The fix is trivial so hopefully @mrook is amenable to applying it to the
current release branch as well as the alpha version.

elyobo:
Coming up on a year later, v3 is still alpha, and this still bites me periodically.  Any interest in applying this trivial fix to improve PHP7 compatibility @mrook @siad007?

mrook:
I think I missed this. I'll take a look at rolling a 2.16.x patch fix with this applied. We don't have any definitive plans for the v3 release roadmap.

elyobo:
:+1: Thanks

PR #981, opened 2018-10-27
Make PHPStan task generate error messages and allow skipping see #980

### LINKED ISSUE #980: PHPStan Task does not fail during errors
**Setup:**
Using the latest master branch 4c47f5a1da93b756e481e818912763cec02e0bfe

**Problem Description**
I am implementing the PHPStan Task like I have found in the Documentation. But it does not fail if an Error occures.

```xml
<phpstan description="Runs phpstan" command="analyse" level="7" paths="src tests"/>
```

**Expected**
The Task sould show an Error Result and Fail.
If I run directly: 
`phpstan analyse --level=7 src tests`
It Fails and showing me the error output like:
```text
------ ------------------------------------------------------------------------------------------------------------------------------
  Line   src/Entity/Item.php
 ------ ------------------------------------------------------------------------------------------------------------------------------
  154    Method Api\Entity\Item::getTags() should return Doctrine\Common\Collections\Collection&iterable<Api\Entity\Item> but
         returns Doctrine\Common\Collections\Collection&iterable<Api\Entity\Tag>.
```

**Debug information**
I tracked it down to `PHPStanTask::main()`. I see the Error inside `return [$return, $output];`. Nut it is not shown.

If you need more information please let me know.

Zeromax:
this will fix #980


Zeromax:
Ok, so I have one question. 
Where do I have to add `checkreturn`? So that it is known in the XML file.

Zeromax:
@mrook I am not 100% sure. In the file `optionaltasks.xml` there was no entry for for phpstan. So I hope I did it right.

mrook:
Looks great! Thanks for your contribution @Zeromax 👍

PR #988, opened 2018-11-11
Update PHPCS to v3

carusogabriel:


carusogabriel:
Looks like this patch is a BC Break, right? :(

siad007:
@carusogabriel thanks for contributing. But there is more work than just bumping up the version in composer. The `PhpCodeSnifferTask` is not compatible with the PHPCodeSniffer v3 API.

carusogabriel:
@siad007 Thanks. I'm gonna update this patch.

Issue #1003, opened 2018-12-05
Feature request: Disable Input


MaximilianKresse:
I want to integrate phing into my Teamcity build process.

Is there is simple way to disable the input? (e.g. in composer there is a "--no-interaction" option)

As far as I can see the only way is to write a own class which implements InputHandler and registers nothing like:
```php
class NoInteractionInputHandler implements InputHandler
{
    public function handleInput(InputRequest $request)
    {
    }
}
```
And call phing with the argument "-inputhandler NoInteractionInputHandler". 

It would be cool if there would be a simple option or a preconfigured InputHandler.

mrook:
Hi @MaximilianKresse, whether phing asks for input is typically dependent on the build file, so I'm not entirely sure this feature makes sense.

MaximilianKresse:
Yeah thats right but I want to be sure that my build process will not run forever if there is a prompt.

Would it hurt to have a "NoInteractionInputHandler"? There isn't much code to care about ;-)

mrook:
Fair enough! No it wouldn't hurt that much ;-) Pull requests welcome!

siad007:
Merged in #1004

PR #1004, opened 2018-12-05
Added DisableInputHandler

MaximilianKresse:
Implementation for #1003 
Just a simple dummy input handler to disable console interaction.


MaximilianKresse:
I added it to the Command Line Arguments documentation page. But a test is a little bit more complicated because it does nothing by itself =)

siad007:
> I added it to the Command Line Arguments documentation page. But a test is a little bit more complicated because it does nothing by itself =)

@MaximilianKresse you could write a test with an `exec` task to call phing as an executable with the new input handler as an argument and an additional buildfile, where you have an input task and an echo task to assert against the echo text in the logs. What do you think?

MaximilianKresse:
I added a test. I have oriented myself to the ConsoleInputHandler. There was a bug because of some unexpected handling of default values (the input handler have to care about the default value).

siad007:
Thanks!

PR #1005, opened 2018-12-06
 Opened IoncubeEncoderTask to all php versions

MaximilianKresse:
Every new php version we come to the problem that the phing task have to manually changed to support the new version. See #335 #485 #130 

To solve this once and for all the phpversion should be removed and allow everything given. So we can also finally use this x64 versions of ioncube encoder without renaming the encoder files.

### DIFF
--- a/classes/phing/tasks/ext/ioncube/IoncubeEncoderTask.php
+++ b/classes/phing/tasks/ext/ioncube/IoncubeEncoderTask.php
@@ -574,12 +574,7 @@ class IoncubeEncoderTask extends Task
     {
         $arguments = $this->constructArguments();
 
-        if (in_array($this->phpVersion, [5, 53, 54, 55, 56, 71])) {
-            $encoderName = $this->encoderName . $this->phpVersion;
-        } else {
-            $encoderName = $this->encoderName;
-        }
-        $encoder = new PhingFile($this->ioncubePath, $encoderName);
+        $encoder = new PhingFile($this->ioncubePath, $this->encoderName . $this->phpVersion);
 
         $this->log("Running ionCube Encoder...");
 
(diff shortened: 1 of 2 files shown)


classes/phing/tasks/ext/ioncube/IoncubeEncoderTask.php
```diff
@@ -574,12 +574,7 @@ public function main()
     {
         $arguments = $this->constructArguments();
 
-        if (in_array($this->phpVersion, [5, 53, 54, 55, 56, 71])) {
-            $encoderName = $this->encoderName . $this->phpVersion;
-        } else {
-            $encoderName = $this->encoderName;
-        }
-        $encoder = new PhingFile($this->ioncubePath, $encoderName);
+        $encoder = new PhingFile($this->ioncubePath, $this->encoderName . $this->phpVersion);
```
siad007: @MaximilianKresse thanks for your contribution. One small suggestion: would you mind to remove the default value of the `$this->phpVersion` with the actual running php version?
```php
// ...
    private $phpVersion;
// ...
        $encoder = new PhingFile(
            $this->ioncubePath,
            $this->encoderName . $this->phpVersion ?? str_replace('.', '', $this->getProject()->getProperty('php.version'))
        );
```
What do you think?

MaximilianKresse: This sounds great but in practice it doesn't work. The variable name "phpVersion" doesn't tell the whole truth - actually its the encoder version. The ioncube encoder has multiple encoder versions and additionally encoders for specific php versions:

grafik
Note that there isn't a php 7 or 70 version so a direct mapping wouldn't work. Not every loader is compatible to every ioncube encoder. 

At the moment we always have to rename the files to match the expected shema of phing. With this change we could leave this ioncube encoder file names and change the "phpVersion" to something like "72_10.2_64".

And additionally I want to have so less logic as possible in this task because we don't know when ioncube likes to change there file structure ;-)

PR #1019, opened 2019-01-29
Feature/visualizer task

jawira:
`VisualizerTask` creates diagrams using buildfiles, these diagrams represents calls and depends among targets.


composer.json
```diff
@@ -26,7 +26,8 @@
         "sebastian/version": "^2.0",
         "symfony/console": "^2.8 || ^3.0 || ^4.0",
         "symfony/yaml": "^2.8 || ^3.1 || ^4.0",
-        "mehr-als-nix/parallel": "^v1.0"
+        "mehr-als-nix/parallel": "^v1.0",
+        "jawira/plantuml-encoding": "^1.0"
```
siad007: The dependency should be moved to the require-dev section for testing.
Could you also add a suggest entry for your lib?

jawira: it should be fixed with my last commit

siad007:
@jawira thanks

PR #1026, opened 2019-02-06
Replaced deprecated create_function calls with anonymous functions.

chasepeeler:
create_function is deprecated in 7.2, but, anonymous functions have been supported since 5.3, so including this update in version 3 (which requires 7.1+) should be harmless.

(I submitted this earlier, but, the commits got ugly and screwed up)


mrook:
Thanks!

Issue #1030, opened 2019-02-10
Bug when using <linecontains> filter on large files


jawira:
When applying `<filtercontains>` on large files, some lines in the resulting file are concatenated. 

To reproduce the bug you need a large file, I generated mine using this script:

```php
<?php
$fileName = 'lines.txt';
unlink($fileName);
$x = 0;
while (++$x <= 10000) {
    file_put_contents($fileName, "This is line $x" . PHP_EOL, FILE_APPEND);
}
```
After executing this script you should have the file `lines.txt` which contains 10000 lines:

```
This is line 1
This is line 2
This is line 3
This is line 4
This is line 5
This is line 6
...
This is line 9997
This is line 9998
This is line 9999
This is line 10000
```

Now to reproduce the bug I created a buildfile like this:

```xml
<?xml version="1.0" encoding="UTF-8" ?>

<project name="lines" default="filter">

    <target name="filter">
        <copy file="lines.txt" tofile="filtered.txt" overwrite="true">
            <filterchain>
                <linecontains>
                    <contains value="0"/>
                </linecontains>
            </filterchain>
        </copy>
    </target>
    
</project>
```

The `filter` target will remove all line which don't contain the `0` string:

```
$ phing filter
```

This will create a new file `filtered.txt`, the problem is that sometimes two lines are concatenated, in this case line `939` will contain two lines from `lines.txt`, this should not happen:

```
...
This is line 3700
This is line 3701
This is line 3702This is line 3703
This is line 3704
This is line 3705
This is line 3706
...
``` 

I know this has something to do with "$buffer" variable, but I don't had enough time to check it in detail. If somebody can give me a pointer I will fix this bug.

siad007:
Hello @jawira 
after a little test, I discovered an other issue regards to the `0` value of `contains`, which somewhere changes to an empty string. But your issue was not reproducable on the latest master branch by myself.
```
<?xml version="1.0"?>
<project name="Phing Build Tests" default="genlist" basedir=".">
    <target name="genlist">
        <php><![CDATA[
            $fileName = 'lines.txt';
            @unlink($fileName);
            $x = 0;
            while (++$x <= 10000) {
                file_put_contents($fileName, "This is line $x" . PHP_EOL, FILE_APPEND);
            }
        ]]></php>

        <copy file="lines.txt" tofile="filtered.txt" overwrite="true">
            <filterchain>
                <linecontains>
                    <contains value="1"/>
                </linecontains>
            </filterchain>
[...]

PR #1031, opened 2019-02-11
Bugfix linecontains

jawira:
Fixed bug defined in issue #1030

Method `\LineContains::read` can be called multiple times when file is too big. When this happens, the last line of the first call is concatenated to first line of the second call. The bug was caused by the lack of trailing "\n" in the return string.

I also fixed `phing-grammer.rng` and added "preferred-install" key in `composer.json` this is supposed to speed up installs since zip files are downloaded instead of cloning repos.


jawira:
I just noticed an error in my code. I'm closing the PR.

PR #1057, opened 2019-04-19
Update IntrospectionHelper.php

pjvdput:
Passing project with method getTag


siad007:
@pjvdput thx for contributing. Would you mind explain what your intention is? As far as I can see there could be trouble if the element is an `UnknownElement` see https://github.com/phingofficial/phing/blob/master/classes/phing/UnknownElement.php#L55

siad007:
Closed as this pr makes no sence to me.

PR #1061, opened 2019-04-26
PHPUnit removed hack

siad007:
Related to https://github.com/sebastianbergmann/phpunit/issues/3611


epdenouden:
@siad007 👍 thanks for the cleanup. Since the logic hasn't been removed completely, I'm wondering what it's used for; something re: incremental flushing?

siad007:
@epdenouden AFAIK this is used for example to disable writing to file by the format element:
```xml
<phpunit>
    <format type="xml" usefile="false"/>
</phpunit>
```
But in fact the whole phpunit task should be refactored...

PR #1130, opened 2019-09-23
add editorconfig, update gitattributes

mimmi20:
This PR want to add add a .editorconfig file and change the gitattributes to make sure that windows users (like me) do not change something unnessary (like line endings).


siad007:
@mimmi20 thx for your contribution!

PR #1129, opened 2019-09-19
Fix coding style

mimmi20:
First I just wanted to add the matrix feature to the .travis.yml file to make it more readable.

Then I had to update the dependencies to get rid of some errors caused by PDepend on PHP 7.2+. After that I updated the coding style and fixed some issues to remove the surpressing of warnings.

Additionally to that I added a editorconfig file and updated the gitattributes to make sure, that I as windows user dont change any lineendings.

PR #1132, opened 2019-09-25
update dependencies

mimmi20:
This PR wants to update the dependencies to avoid some errors caused from PDepend on PHP 7.2+.


siad007:
@mimmi20 thank you 😃

PR #1135, opened 2019-09-29
FileHashTask should always generate a file.

siad007:
We should always create a file.
Instead of:
```xml
<filehash file="${filename}" algorithm="sha512" propertyname="phar.hash"/>
<basename property="basefilename" file="${filename}"/>
<echo file="${filename}.sha512">${phar.hash} ${basefilename}</echo>
```
just
```xml
<filehash file="${filename}" algorithm="sha512" />
```


mrook:
Agreed!

PR #1136, opened 2019-09-30
add Build-Matrix

mimmi20:
This PR wants to add a Build-Matrix to make the travis file more readable and group the tests.

As a result
- as first step the composer.json and composer.lock are validated
- as second step the coding style is checked
- for each PHP version the project is installed with lowest and normal dependency versions
- for each PHP version PHPUnit is run without XDebug to make the tests faster
- an additional test is run on the nightly build of PHP without dependency limits
- as last step PHPUnit is run with XDebug

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -2,6 +2,17 @@ language: php
 
 sudo: false
 
+php:
+  - 7.1
+  - 7.2
+  - 7.3
+  - 7.4snapshot
+  - master
+
+env:
+  - DEPS="LOW"
+  - DEPS="NORMAL"
+
 addons:
   apt:
     packages:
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -12,58 +23,89 @@ cache:
...
+      php: 7.2
+      env: DEPS="NORMAL"
+      script:
+        - composer validate
+
+    - stage: static code analysis
+      os: linux
+      php: 7.2
+      env: DEPS="NORMAL"
+      script: bin/phpcs -s --standard=./ruleset.xml classes
```
siad007: Please add [`--config-set ignore_warnings_on_exit 1`](https://github.com/squizlabs/PHP_CodeSniffer/wiki/Configuration-Options#ignoring-warnings-when-generating-the-exit-code) to avoid the build to fail on warnings.

mimmi20: I would prefer the `-n` option as it was before.

siad007:
@mimmi20 thanks!

PR #1137, opened 2019-09-30
update coding style

mimmi20:
This PR wants to update the coding style. All surpressed warning in the coding style are fixed.


ruleset.xml
```diff
@@ -1,11 +1,20 @@
 <?xml version="1.0"?>
 <ruleset name="phing">
     <description>
-        PSR12 style with no namespace.
+        PSR2 style with no namespace.
```
siad007: Please use psr12 here

ruleset.xml
```diff
@@ -1,11 +1,20 @@
...
 
-    <rule ref="PSR12">
+    <!-- PHP keywords MUST be in lower case. -->
+    <rule ref="Generic.PHP.LowerCaseKeyword"/>
+
+    <rule ref="PSR2">
         <exclude name="PSR1.Classes.ClassDeclaration.MissingNamespace"/>
         <exclude name="PSR2.Methods.MethodDeclaration.Underscore"/>
+        <exclude name="PSR2.Classes.PropertyDeclaration.Underscore"/>
+        <exclude name="Generic.Files.LineLength.TooLong"/><!-- can not be suppressed -->
```
siad007: Could you remove the excludes? It just will hide all stuff which should be fixed instead.

ruleset.xml
mimmi20: fixed

ruleset.xml
mimmi20: fixed (expect the last 2, see the comment behind the excludes)

PR #1086, opened 2019-07-02
phing wouldn't work from the test directory, document how I got it to work

kenguest:
phing kept throwing an error, as shown below, when attempting to run it from the test directory.
got it to run by doing this from root directory instead...
```
$ ./bin/phing -f test/build.xml -Dtest=NotifyTaskTest

Phing Build Tests > initialize:


Phing Build Tests > configure:

     [echo] -------------------------------------------------
     [echo]  +++++ Running Phing  unit tests
     [echo] -------------------------------------------------

Phing Build Tests > reports:

   [delete] Deleting directory /home/kguest/dev/code/phing/test/tmp
    [mkdir] Created dir: /home/kguest/dev/code/phing/test/tmp
#!/usr/bin/env php
```

PR #1143, opened 2019-10-02
Rename __import method

kenguest:
Rename __import method as PHP has reserved all method names with a double underscore prefix for future use and this causes a conflict


README.md
```diff
@@ -123,6 +123,10 @@ Then, perform the following steps (on a clone/fork of Phing):
          $ cd test
          $ ../bin/phing
 
+Should this not work, you might be able to run the unit tests from the main directory:
+
+         $ $ ./bin/phing -f test/build.xml -Dtest=NotifyTaskTest
+
```
siad007: @kenguest This is not needed. see #1086

kenguest: My bad - that was a straggler from a previous PR, removed it now.

PR #1144, opened 2019-10-02
    The variable '$php_errormsg' is deprecated since PHP 7.2; Using error_get_last() instead

kenguest:
The variable '$php_errormsg' is deprecated since PHP 7.2; Using error_get_last() instead
    
Also using error_clear_last (Available since PHP 7) to clear any previously set error message
to ensure only relevant error message is referenced/used.


classes/phing/system/io/FileSystem.php
```diff
@@ -293,13 +298,18 @@ public function getLastModifiedTime(PhingFile $f)
      */
     public function getLength(PhingFile $f)
     {
+        if (PHP_MAJOR_VERSION >= 7) {
```
siad007: I think we could drop all occurrences of the `PHP_MAJOR_VERSION >= 7` conditions as the minimum php version requirement for phing is PHP >= 7

kenguest: I've dropped that check on PHP_MAJOR_VERSION now.

PR #1176, opened 2019-10-25
Parameter unittests

kenguest:


kenguest:
I've renamed the test so it should be ok to merge now.

PR #1187, opened 2019-11-09
[wip] Fix tests

mimmi20:


classes/phing/filters/TailFilter.php
```diff
@@ -85,6 +85,9 @@ public function read($len = null)
         }
 
         while (($buffer = $this->in->read($len)) !== -1) {
+            // Replace "\r" and "\r\n" with "\n"
+            $buffer = str_replace(["\r\n", "\r"], "\n", $buffer);
```
siad007: Instead of struggling with line endings, we should go by `$this->readLine()` as already implemented in `classes/phing/filters/LineContains.php` line 110. Therefor we have to rewrite the whole method.

mimmi20:
> Sry but I do not see benefit of using `self::assert*` instead of `$this->assert*` ! Would you mind explain your changes here?

These functions are defined as `static`.

siad007:
@mimmi20 please split up this pr in several ones as there are mixed up implementation fixes and test fixes.

mimmi20:
ok

siad007:
> 
> 
> > Sry but I do not see benefit of using `self::assert*` instead of `$this->assert*` ! Would you mind explain your changes here?
> 
> These functions are defined as `static`.

Yes you are right, there are defined static, but they are used non static everywhere and also documented as such in https://phpunit.readthedocs.io/en/8.4/assertions.html

PR #1188, opened 2019-11-09
fix coding style for test files

mimmi20:


test/classes/example/tasks/TestEcho.php
```diff
@@ -0,0 +1,30 @@
+<?php
+
+/*
```
siad007: Would you mind changing the file comments to `/**` and omit the empty line 2.

mimmi20: fixed

siad007: > 
> 
> fixed

no

mimmi20: I found more files with this type of doc block and fixed them all. I added this doc block to files where it was missing.

siad007:
@mimmi20 thx 👍

PR #1212, opened 2019-12-17
[WIP] extend Coding Style

mimmi20:
This PR is still work in progress.


kenguest:
What exactly is this about? I don't believe there's any intent on changing the Phing codebase from established PSRn Coding Standards... Please correct me if I'm wrong.

mrook:
You're not wrong @kenguest, there's no intent to change that. Can you clarify @mimmi20 ?

mimmi20:
The Doctrine coding standard and the underlying Slevomat coding standard are based on PSR 2. I want to use this to force strict types, type hints (as much as possible), doc block formating and single quotes.

Because PSR 2 is different from PSR 12, there are some rule conflicts I have to solve. After this I want to play around with the rules a bit.

siad007:
IMO it makes no sense to downgrade from PSR12 to PSR2

Thomas Müller <notifications@github.com> schrieb am Di. 17. Dez. 2019 um
17:29:

PR #1244, opened 2020-01-27
Provide --config-option switch for svn tasks

MiRacLe-RPZ:
In some cases i need set _--config-option_ switch for svn commands (ex. _--config-option config:miscellany:use-commit-times=yes_). This PR add configOption property for all svn tasks.

### DIFF
--- a/docs/guide/en/source/appendixes/optionaltasks.xml
+++ b/docs/guide/en/source/appendixes/optionaltasks.xml
@@ -12081,7 +12081,7 @@ host="webserver" command="ls" /></programlisting>
                         <entry><literal>configOption</literal></entry>
                         <entry><literal role="type">String</literal></entry>
                         <entry>Override subversion's config option</entry>
-                        <entry><literal>false</literal></entry>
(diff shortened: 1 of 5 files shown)


docs/guide/en/source/appendixes/optionaltasks.xml
```diff
@@ -12077,6 +12077,13 @@ host="webserver" command="ls" /></programlisting>
                         <entry><literal>false</literal></entry>
                         <entry>No</entry>
                     </row>
+                    <row>
+                        <entry><literal>configOption</literal></entry>
+                        <entry><literal role="type">String</literal></entry>
+                        <entry>Override subversion's config option</entry>
+                        <entry><literal>false</literal></entry>
```
siad007: Please change `false` to `n/a` as it could suggest a bool type. (same for all other occurrencies)
Also a small example would be good to have.
Is there also any change to have a unit test for it?

siad007:
@MiRacLe-RPZ Thanks 👍

PR #1247, opened 2020-01-28
Simplify visualizer tests

jawira:
In order to avoid errors in VisualizerTask tests, I removed any reference to any external PlantUML server.


siad007:
@jawira thx

PR #1288, opened 2020-03-26
Replace HTTP_Request2 with GuzzleHttp

### LINKED ISSUE #1160: get_magic_quotes_runtime() deprecated in 7.4 - replace HTTP_Request2
`HTTP_Request2`  still uses `get_magic_quotes_runtime()` and hasn't seen a real release since 2016. We should probably replace this package with something more modern & lightweight.

mrook:
Fixes #1160.

What do you think @siad007 @kenguest ?


mrook:
Docs still need to be updated.

PR #1294, opened 2020-03-29
[WIP] [PhingTask] Fixed possible infinity loop

siad007:
Fixed infinity loop if a task calling a target that depends on its parent target.

Related to #1241


siad007:
Closed in favor of finished and functional #1241

PR #1393, opened 2020-09-18
Variable properties which could not expanded should be null.

### LINKED ISSUE #1391: Error resolving proxy
Hi, I had the habit to add proxy configuration to my buildfiles. Because in linux you can set your proxy using an environment variable I simply passed this variable to a property:

```xml
<property name="phing.http.proxy" value="${env.http_proxy}"/>
```

Currently I don't work behind a proxy anymore, so `${env.http_proxy}` doesn't exists. This wasn't a problem until I updated to **Phing 3.0.0-alpha4**:

```console
$ phing -v
Phing 3.0.0-alpha4

$ phing emoji-list:download
Buildfile: /home/jawira/PhpstormProjects/emoji-catalog/build.xml

Emoji catalog > emoji-list:download:


BUILD FAILED
/home/jawira/PhpstormProjects/emoji-catalog/build.xml:68:98 cURL error 5: Could not resolve proxy: ${env.http_proxy} (see https://curl.haxx.se/libcurl/c/libcurl-errors.html)

Total time: 0.0630 seconds
```

Apparently now `phing.http.proxy` must be a valid proxy if it's declared. I understand that this could be related to the use of Guzzle library, but do you thing is it possible to revert to the previous behavior?

siad007:
After we discovered that we have an expandable variable, we give it back unchecked. We should return null in case of having a not expandable property.

Fixes #1391

### DIFF
--- a/classes/phing/PropertyHelper.php
+++ b/classes/phing/PropertyHelper.php
@@ -428,6 +428,9 @@ class PropertyHelper
         if (false !== strpos($found, '${')) {
             // attempt to resolve properties
             $found = $this->replaceProperties($found, null);
+            if (StringHelper::startsWith('${', $found) && StringHelper::endsWith('}', $found)) {
+                $found = null;
+            }
             // save resolved value
             $this->properties[$name] = $found;
         }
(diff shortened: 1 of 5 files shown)


jawira (approved):

jawira:
I tested with my code and it works, thank you 🎉

PR #1396, opened 2020-09-22
Fix Support for Nested Exec Env Tag - Fixes #1395

### LINKED ISSUE #1395: ENV SubTask Not Working as Expected
For the `env` sub-task of the `exec` task, [this line](https://github.com/phingofficial/phing/pull/1254/files#diff-3a70b8da0bfb1d87d650373b9e443af7R274) has the effect of creating multiple, distinct commands - not the intended effect of adding an environment variable to a command.  This works for simple commands like `echo` (as per the test), but does not work as expected for scripts that delegate execution to other commands.  For example, this does not work when executing `./vendor/bin/behat`, which ultimately calls through to `php` where the environment variables are then lost.

It does not work on Linux or Mac operating systems.  Would there be a downside to removing the the semicolon?

jimdelois:
Fixes #1395 .  See linked issue.


siad007 (changes requested):
Thanks for your contribution.
Please add an additional test.

jimdelois:
@siad007 - The original `testEnvVars` test covered, but I added an additional test with a new command that also ensures that multiple env values are correctly processed.

test/etc/tasks/system/ExecTest.xml
```diff
@@ -206,12 +206,21 @@
...
         </exec>
         <echo msg="${hello}"/>
     </target>
 
+    <target name="testMultipleEnvVars">
+        <exec executable="php" outputProperty="outputProperty">
+            <env key="HELLO" value="hello"/>
+            <env key="WORLD" value="world"/>
+            <arg value="-r"/>
+            <arg value="print '$HELLO ' . '$WORLD';"/>
```
siad007, line 222: This seems a little bit strange to me - imagine one of the variables are used inside the code itself.

siad007 (changes requested):

test/etc/tasks/system/ExecTest.xml
jimdelois: Right, that would be a more conclusive assertion.  Updated.

siad007 (changes requested):
Some tests are failing...

jimdelois:
Tests are failing in `PDOSQLExecTaskConditionTest` which have no references to `<env />` and seems to be unrelated to my change (also it passed the previous build(s)).  How does one go about rekicking a build?

PR #1419, opened 2020-10-15
Fixed env handling for windows

siad007:
* Fixed env handling for windows
* Added test case for windows


mrook:
This fixes #1395.

PR #1420, opened 2020-10-19
Add unit attribute to FileSizeTask

jawira:
I added `unit` attribute to `FileSizeTask`. This allow to convert the size from bytes to another unit on the fly.

For example:

```xml
<target name="backup:size">
    <filesize file="./backup.zip" propertyname="backup.size" unit="M"/>
    <echo>Your backup size is ${backup.size} MB</echo>
</target>
```

Possible value for `unit` are: B, K, M, G, T, P.

I chose these values to be similar as possible to values used by other elements:

- `TrucateTask`: k, m, g.
- `size` selector: k, m, g, t. ([and a LOT more](https://github.com/phingofficial/phing/blob/2a6e37599053bbec23c22b6efb1c2d95cbf9ecdc/classes/phing/types/selectors/SizeSelector.php#L62))
- `hasfreespace` condition: K, M, G, T, P

As usual I'm open to any comment or question.


jawira:
PS: My IDE become crazy and changed phing.topleveltest.xml, I don't know what happened since the content is the same. Sorry about that.

jawira:
PPS: can you tell me where documentation is please? I can't found it.

mrook:
> PS: My IDE become crazy and changed phing.topleveltest.xml, I don't know what happened since the content is the same. Sorry about that.

Probably newlines?

mrook:
> PPS: can you tell me where documentation is please? I can't found it.

https://github.com/phingofficial/guide

PR #1427, opened 2020-11-08
ComposerTask tests don't need Composer on the path

### LINKED ISSUE #1426: ComposerTask test failing if Composer isn't on the path
If Composer isn't on the path the testSetGetComposer method will fail.

Test should simulate composer on the path instead of requiring Composer be installed and on the path.

BlairCooper:
Fixes #1426

Split out testing SetComposer and GetComposer methods. Simulate Composer
existing on the path.

### DIFF
--- a/test/classes/phing/tasks/ext/ComposerTaskTest.php
+++ b/test/classes/phing/tasks/ext/ComposerTaskTest.php
@@ -88,16 +88,24 @@ class ComposerTaskTest extends \PHPUnit\Framework\TestCase
     /**
      * @covers ComposerTask::getComposer
      */
-    public function testGetComposer()
+    public function testGetComposerNotOnPath()
     {
         $composer = 'bar';
         $o = $this->object;
 
+        $orgPath = getenv("PATH");
+        
         $prop = new ReflectionProperty('ComposerTask', 'composer');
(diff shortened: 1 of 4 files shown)


test/classes/phing/tasks/ext/ComposerTaskTest.php
```diff
@@ -72,18 +72,63 @@ public function testSetGetPhp()
...
+
+        $testPath = PHING_TEST_BASE . '/etc/tasks/ext/composer';
+        $orgPath = getenv("PATH");
+
+        $pathSeparator = FileSystem::getFileSystem()->getPathSeparator();
+        putenv("PATH=$testPath$pathSeparator$orgPath");
+
+        $pathComposer = $o->getComposer();
+
+        putenv("PATH=$orgPath");
```
siad007, line 128: AFAIK windows uses `Path` and not `PATH` 🤔 ?

BlairCooper: Windows isn't case sensitive so Path, PATH, and pAtH would all work.

siad007:
@BlairCooper https://github.com/phingofficial/phing/issues/1426#issuecomment-723648918

PR #1436, opened 2020-11-15
Add new optional clover-html formatter.

### LINKED ISSUE #1430: Make better use of PHPUnit 9 code coverage and reporting
- Pass phpunit configuration file values onto PHPUnit filter and code coverage classes
- Instead of relying on the current coverage-* tasks, take advantage of the code coverage reporting (HTML)

BlairCooper:
Fixes #1430

~~Add runner for PHPUnit 9~~
Add support for Clover HTML report.
Pass PHPUnit configuration values to filter and code coverage classes.


siad007 (changes requested):
Please revert the changes from the runner or do it complete - again there are missing stuff!
FYI I have fixed the coverage issue in #1437

Issue #1449, opened 2020-11-26
DateSelector & TouchTask should use seconds instead of milliseconds


BlairCooper:
With the default granularity in PHP being seconds, DateSelector and TouchTask should also use seconds instead of milliseconds.

Note: the classes do use seconds but the interface and documentation suggest that it uses milliseconds.

siad007:
As described in the documentation the [DateSelector](https://www.phing.info/guide/hlhtml/#Date) uses seconds.
But there might be a hidden feature (as seen in the code) to support milliseconds (like ANT does) in the future.
The [TouchTask](https://www.phing.info/guide/hlhtml/#TouchTask) uses milliseconds as a creation date starting from the unix epoche. So this line:
```xml
<touch file="README.txt" millis="102134111" />
```
will create the file with a timestamp of 102134111 milliseconds after midnight Jan 1 1970

What exactly is your expectation here?

siad007:
@BlairCooper please add an example/test which fails, to understand your intention.

siad007:
Issue splitted.
Related to #1452 #1453 #1454

PR #1450, opened 2020-11-26
Change TouchTask and DateSelector to use seconds

BlairCooper:
Fixes: #1449

Changed code for TouchTask and DateSelector to be based on seconds instead of milliseconds.

Corrected tests to use appropriate test date format compatible with strtotime.


siad007:
@BlairCooper would you mind to separate into three PRs please.
1. code style fixes
2. DateSelector
3. TouchTask
Also do **not** deprecate `millis`. phing is a fork of ANT and should use the API as much as possible.

BlairCooper:
Reworking

PR #1457, opened 2020-11-29
Changed arg passed to phploc

BlairCooper:
Changed argument passed to phploc in testNestedEnv from -V to -v. phploc
7.x doesn't support -V anymore.


siad007:
@BlairCooper thx

PR #1472, opened 2020-12-08
Only notify to Slack for phingofficial builds

BlairCooper:


.github/workflows/build.yml
```diff
@@ -2,6 +2,8 @@ name: Phing CI
 
 on:
   push:
+    branches:
+      - master
```
mfn: Curious, what made your mind change?

BlairCooper: Was done to test changes to notify job on forked repository. This put it back to the way it was.

siad007: The actions were triggered twice on each PR.
One for push and one for the pull request each.
Do you have ideas to configure it to not trigger the actions twice on pull request level?

BlairCooper: Assuming the lines above are not present here is what I've seen/believe should happen.
- When I push to my forked repo the build runs against my repo/branch, nothing happens on phingofficial (push trigger)
- If I push again to my repo the build runs against my repo/branch, nothing happens on phingofficial (push trigger)
- When I submit a pull request, the build runs on phingofficial/mybranch (pull_request/opened trigger)
- If I push again to my repo, the build runs in my repo/branch (push trigger) and in phingofficial/mybranch (pull_request/synchronize trigger)
- When the pull request is accepted into master a build would run in phingofficial/master (push trigger)

Are you seeing something else? I'm guessing you're working against the phingofficial repo directly, no fork.
If that's the case I would expect you to see a build on your branch each time you push (push trigger).
And when you merge into master (push trigger).
You shouldn't see any pull request triggers as you aren't creating pull requests for your own changes.

siad007: >  I'm guessing you're working against the phingofficial repo directly, no fork.
Your guess is wrong I am using a fork in between :)
The problem here is that if you make a PR from a fork then the action is executed once but
after pushing again then there are two executions one for the change inside the pr and one for the push itself.
More descriptive info could be found [here](https://github.community/t/how-to-trigger-an-action-on-push-or-pull-request-but-not-both/16662) Will have a look through this post and change it then.

BlairCooper: But when there are two executions isn't one on the source repo/fork (push trigger) and one on target repo/phingofficial (pull_request synchronize)? A bit repetitive as they'll both have the same result.

BlairCooper: With this, it would be the source repo build that's somewhat extraneous. You would want the target repo build to run so that the owners can see when the build is succeeding and subsequently merge the pull request. It might be nice if there was something that would only trigger the target repo build when the source repo had built successfully. I think these builds each run in their own little world so they don't know that another repo is also building the push.

BlairCooper: And are you forking, or branching? Looking at siad007-patch-4 for example, this was a branch in phingofficial, not a branch in a forked repo. If you then did a pull request to merge siad007-patch-4 into master I would expect that you'd get two builds in the same repo.

siad007: Fixed with https://github.com/phingofficial/phing/pull/1474

PR #1477, opened 2020-12-14
Show suggestion on unknown target

jawira:
This adds the message `Did you mean 'target'?` when target is not found.

For example:

```xml
<?xml version="1.0" encoding="UTF-8" ?>

<project name="demo" default="hello">
    <target name="setup">
        <echo>This is setup</echo>
    </target>
    <target name="hello">
        <echo>This is hello</echo>
    </target>
    <target name="clean">
        <echo>This is clean</echo>
    </target>
</project>
```

```console
$ bin/phing hell
Buildfile: /home/jawira/PhpstormProjects/fork-phing/build.xml

BUILD FAILED
 Target 'hell' does not exist in this project. Did you mean 'hello'?

Total time: 0.1012 seconds
```

This was some kind of pet feature for me, so don't hesitate to close this PR if you think it's not useful :)

PR #1478, opened 2020-12-15
[core] Fixed codecoverage in action with latest ubuntu

siad007:
Related to #1477


siad007:
This was made in the origin PR #1477

PR #1482, opened 2020-12-20
Size helper

### LINKED ISSUE #1480: Standardize file size units
Many tasks transforms Bytes to another unit, and from any unit to Bytes.

- TrucateTask: k, m, 
- size selector: k, m, g, t and many variations.
- hasfreespace condition: K, M, G, T, P
- FileSizeTask: B, K, M, G, T, P.

I have already mentioned this in #1420

Every mentioned Task handles the unit conversion by itself (except `TrucateTask` which uses `\Phing::convertShorthand`), so I want to create `SizeHelper` class, the idea is to put all the conversion logic in a single place (the same way `StringHelper` is used to parse strings).

1024 is used as multiplier almost everywhere, except in `size selector` (which also supports 1000 as multiplier). Therefore -in order to apply KISS philosophy- I think the best is to only multiply bytes by 1024 and unsupport 1000. I know this is a breaking change, but that why major releases exist right?  

What do you think? Should I continue with this idea and create a PR?

jawira:
Closes #1480 

- [x] Create SizeHelper
- [x] Refactor HasFreeSpaceCondition
- [x] Refactor FileSizeTask
- [x] Refactor SizeSelector
- [x] Refactor TrucateTask
- [x] Delete Phing::convertShorthand()


jawira:
Breaking changes:

- `SizeSelector` only has two attributes now, "values" and "when". The "units" attributes is not valid any more, however now you can append the unit at the end of "values". 
    ```xml
    <!-- before -->
    <size value="2" when="more" units="g"/>
    <!-- after -->
    <size value="2g" when="more"/>
    ```
- Also `SizeSelector` only supports 1024 as multiplier value.
- I don't know why but now `TruncateTask` returns float instead of int. See tests `test/classes/phing/tasks/system/TruncateTaskTest.php`

siad007:
> 
> 
> Breaking changes:
> 
>     * `SizeSelector` only has two attributes now, "values" and "when". The "units" attributes is not valid any more, however now you can append the unit at the end of "values".
>       ```
>       
>       <size value="2" when="more" units="g"/>
>       
>       <size value="2g" when="more"/>
>       ```

IMO it would be better to have it separated for better dealing with properties here. For later usage of the real value you have now to get rid of the unit.

>     * Also `SizeSelector` only supports 1024 as multiplier value.

😢 we should also support IEC standard notation as well.

>     * I don't know why but now `TruncateTask` returns float instead of int. See tests `test/classes/phing/tasks/system/TruncateTaskTest.php`

`Phing::convertShorthand($adjust): int;` vs `SizeHelper::fromHumanToBytes($adjust): float;`

My suggestion would be to extend the `SizeHelper::fromHumanToBytes` method to `SizeHelper::fromHumanToBytes($adjust, $binaryPrefix = false, $getAsFloat = false): int|float;`

jawira:
> > Breaking changes:
> > ```
> > * `SizeSelector` only has two attributes now, "values" and "when". The "units" attributes is not valid any more, however now you can append the unit at the end of "values".
> >   ```
> >   
> >   <size value="2" when="more" units="g"/>
> >   
> >   <size value="2g" when="more"/>
> >   ```
> > ```
> 
> IMO it would be better to have it separated for better dealing with properties later in the build script. For later usage of the real value you have now to get rid of the unit.
> 
> > ```
> > * Also `SizeSelector` only supports 1024 as multiplier value.
> > ```
> 
> cry we should also support IEC standard notation as well.


Yes, but the thing is `SizeSelector` is the only element that behaves like that. Other elements `TrucateTask`, `HasFreeSpaceCondition` and `FileSizeTask` uses _1024_. That's why I called issue #1480 "**Standardize** file size units". 

The same applies for the "human" notation like `2g`, `SizeSelector` is the exception because it is the only element that has separated attributes, `TruncateTask` and `HasFreeSpace` had always handled this "human" notation to convert sizes to bytes.

> > ```
> > * I don't know why but now `TruncateTask` returns float instead of int. See tests `test/classes/phing/tasks/system/TruncateTaskTest.php`
> > ```
> 
> `Phing::convertShorthand($adjust): int;` vs `SizeHelper::fromHumanToBytes($adjust): float;`
> 
> My suggestion would be to extend the
> `SizeHelper::fromHumanToBytes` method to
> `SizeHelper::fromHumanToBytes($adjust, $binaryPrefix = false, $getAsFloat = false): int|float;`

Well, no. According to my code this is the signature:

```php
public static function fromHumanToBytes(string $human): int
```

Anyway, maybe it would be a good idea to return a _float_ and this way not lose precision.

classes/phing/types/selectors/SizeSelector.php
```diff
@@ -28,186 +28,60 @@
...
-     * This function translates all the long and short forms that a
-     * unit prefix can occur in and translates them into a single
-     * multiplier.
-     *
-     * @param array $units The units to compare the size to.
+     * @param string $value
      *
      * @return void
      */
-    public function setUnits($units)
```
siad007, line 166: Please revert this function. Mix different things is not that good.
Why should a `value` know about its unit? This will make build scripts more complex (having properties twice one with and one without unit)

classes/phing/types/selectors/SizeSelector.php
```diff
@@ -28,186 +28,60 @@
...
-
-    /**
-     * @var array $sizeComparisons
-     */
-    private static $sizeComparisons = ["less", "more", "equal"];
-
-    /**
-     * @var array $byteUnits
-     */
-    private static $byteUnits = [
```
siad007, line 62: Loosing features is not the best way. Please leave it or think about implementing this for the others as well.

classes/phing/Phing.php
```diff
@@ -1719,7 +1688,7 @@ private static function setIni(): void
         self::$origIniSettings['short_open_tag'] = ini_set('short_open_tag', 'off');
         self::$origIniSettings['default_charset'] = ini_set('default_charset', 'iso-8859-1');
 
-        $mem_limit = (int) self::convertShorthand(ini_get('memory_limit'));
+        $mem_limit = intval(SizeHelper::fromHumanToBytes(ini_get('memory_limit')));
```
siad007, line 1691: Please use type cast instead of a function.

classes/phing/util/SizeHelper.php
```diff
@@ -0,0 +1,85 @@
+<?php
+declare(strict_types=1);
```
siad007, line 2: Please add the license file comment.

test/classes/phing/system/util/SizeHelperTest.php
```diff
@@ -0,0 +1,158 @@
+<?php
+declare(strict_types=1);
```
siad007, line 2: Please add file comment.

test/classes/phing/tasks/system/TruncateTaskTest.php
```diff
@@ -43,33 +43,33 @@ public function tearDown(): void
     public function testBasic()
```
siad007, line 43: Please revert this test file and fix the return type. Should be int.

classes/phing/types/selectors/SizeSelector.php
```diff
@@ -28,186 +28,60 @@
...
-            $buf .= "less";
-        } elseif ($this->cmp === 1) {
-            $buf .= "more";
-        } else {
-            $buf .= "equal";
-        }
-        $buf .= "}";
-
-        return $buf;
+        $format = '{%s value: %s when: %s}';
```
siad007, line 53: `$format = '{%s value: %s compare: %s}';`

siad007 (changes requested):

classes/phing/types/selectors/SizeSelector.php
jawira: Mixing _1000_ and  _1024_  is confusing, specially because other tasks only support _1024_.

This is exactly why I created an issue first (#1480 ), I explained everything before starting coding.
A new major version will be released and this is exactly when breaking changes are done.

mrook:
I don't know off the top of my head, does Ant have a convention in this area?

siad007:
No, ant does not have a convention here. It was made by ant exacly how our current state is.
IMO all could have the same behavior but not by loosing features. So my suggestion is to have binary prefix also supported by TruncateTask , HasFreeSpaceCondition ... 
I will merge and apply that missing things in a later PR?! WDYT @mrook

jawira:
Ok, clearly we have different visions here.

Since I only removed features -but don't added any- **nothing will be lost** if I close this PR.
In future issues I will try to be more clear about my intentions.

No hard feelings 👍🙂

mrook:
@jawira I think you may have misjudged mine and @siad007's comments, I don't think the intention was to get you to close the PR/issue :-) Yes, because we are releasing a new major we have a chance to clean up / make things more consistent, but that would warrant some more discussion, don't you think?

jawira:
This PR was only refactoring, no new feature. But the whole point was simplifying and standardize things.

You are doing great job with Phing, I Wish you a merry Christmas guys 🔔🎉🎅🎁🎄 .

siad007:
All a merry Christmas also from me. Please please reopen this PR - it is
the right way to go and I would like to jump in here... 🎄♥️
Jawira Portugal <notifications@github.com> schrieb am Do. 24. Dez. 2020 um
17:27:

jawira:
Sorry if I misinterpreted your words :)

Ok guys, I will implement support for SI and IEC suffixes. 
Are you ok with the following proposition? Or do you want I implement ALL the current prefixes?


| Multiplier     |  Suffix (not case-sensitive)          | 
|-----------------|------------------------------------------------| 
|  1024  | K, M, G, T, P <br> Ki*, Mi*, Gi*, Ti*, Pi*   | 
|  1000  | KB*, MB*, GB*, TB*, PB*                         |

I will wait for your answer, also please give me one or two weeks to finish this.

mrook:
I'm impartial to Ki/Mi/Gi/Ti/Pi (not sure I've ever used them personally, tbh), but the other suffixes are very useful. WDYT @siad007 ?

siad007:
In general I would go with:

| Multiplier     |  Suffix (not case-sensitive)          | 
|-----------------|------------------------------------------------| 
|  1024  | K, M, G, T, P | 
|  1000  | Ki*, Mi*, Gi*, Ti*, Pi* |

With this we could determine the multiplier by char count of the unit and we do not care about `B` as this is the base.
Also we could easily extend by adding more aliases to this list (see SizeSelector)

Bytes should be natural numbers so I would prefer to use integer and only for logging reasons I would make it like `1.456 MB`

Later I also would like to see
* `unit` as an optional param for TruncateTask and HasFreeSpaceCondition
* `value` with or without unit (1M / 1000) which has precedence over a given unit param (?)

jawira:
Ok guys, my bad. I have been doing some research and is not as easy as I put it in the previous table. 
The reason is that thing like "Mi*" (with a wildcard) is not going to work since the full word for "MiB" is "mebibyte", as you can see the second letter is already different.

Therefore I think the best is support the suffixes without wildcards, as the current implementation of SizeSelector does.

In order to be as standard as possible, I have based this proposal on <https://en.wikipedia.org/wiki/Byte#Multiple-byte_units>.

| Multiplier      |   Suffix (not case-sensitive)     |
| --------------  | -------------------------------   |
|  1024 (IEC)     | Ki, KiB, kibi, kibibyte<br>Mi, MiB, mebi, mebibyte<br>Gi, GiB, gibi, gibibyte<br>Ti, TiB, tibi, tebibyte |
|  1000 (SI)  | k, kB, kilo, kilobyte<br>M, MB, mega, megabyte<br>G, GB, giga, gigabyte<br>T, TB, tera, terabyte |

siad007:
@jawira awsome! @mrook your opinion?

siad007:
@jawira @mrook one more thing what came to my mind was the [shorthand bytes](https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes) PHP uses in `php.ini`
Maybe we should move K, M, G, T better to the IEC group?!

mrook:
I like the matrix @siad007 posted a few messages above this one:


Multiplier | Suffix (not case-sensitive)
-- | --
1024 | K, M, G, T, PKi*, Mi*, Gi*, Ti*, Pi*
1000 | KB*, MB*, GB*, TB*, PB*

jawira:
> @jawira @mrook one more thing what came to my mind was the [shorthand bytes](https://www.php.net/manual/en/faq.using.php#faq.using.shorthandbytes) PHP uses in `php.ini`
> Maybe we should move K, M, G, T better to the IEC group?!

Thanks for adding that link, that convinced me :) 

Screenshot from 2021-01-05 15-34-10


| Multiplier      |   Suffix (case-insensitive)     |
| --------------  | -------------------------------   |
|  1024 (IEC)     | K, Ki, KiB, kibi, kibibyte<br>M, Mi, MiB, mebi, mebibyte<br>G, Gi, GiB, gibi, gibibyte<br>T, Ti, TiB, tibi, tebibyte |
|  1000 (SI)  | kB, kilo, kilobyte<br>MB, mega, megabyte<br>GB, giga, gigabyte<br>TB, tera, terabyte |

Is everybody ok with this previous table?

mrook:
Yep, let's roll!

classes/phing/types/selectors/SizeSelector.php
jawira: Done, all these units are supported

classes/phing/Phing.php
jawira: Done

classes/phing/util/SizeHelper.php
jawira: Done

test/classes/phing/system/util/SizeHelperTest.php
jawira: Done

classes/phing/types/selectors/SizeSelector.php
jawira: Done

siad007:
Hi @jawira - great work here!
Is there a reason to not return an integer for `public static function fromHumanToBytes(string $human): float`?
Please also note that this PR is still marked as `Draft`.

jawira:
> Hi @jawira - great work here!
> Is there a reason to not return an integer for `public static function fromHumanToBytes(string $human): float`?
> Please also note that this PR is still marked as `Draft`.

I must confess I'm pretty happy with my code :)

About the `float` value, I have to reasons:

1. The main reason is that the result is not always an int. For example: `2.3K` will return `2355.2` in bytes.
2. Also, PHP `filesize()` function returns a `float` (I don't know how is this possible, I only read the doc :P), which ultimately is the function that `SizeSelector` uses. So, in order to compare `float` with `float` I made `SizeHelper::fromHumanToBytes` to return a `float` as well.

I will remove the draft status.

siad007 (approved):

PR #1501, opened 2021-01-18
[php 8.0 compat] fix php warning:  Private methods cannot be final

Alexander1000:
Private methods cannot be final as they are never overridden by other classes

### DIFF
--- a/classes/phing/listener/AnsiColorLogger.php
+++ b/classes/phing/listener/AnsiColorLogger.php
@@ -157,7 +157,7 @@ class AnsiColorLogger extends DefaultLogger
      * Set the colors to use from a property file specified by the
      * special ant property ant.logger.defaults
      */
-    final private function setColors()
+    private function setColors()
     {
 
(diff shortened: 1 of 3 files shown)


Alexander1000:
@mrook @siad007 review please

siad007:
And in master https://github.com/phingofficial/phing/blob/master/classes/phing/listener/AnsiColorLogger.php

Alexander1000:
need to 2.16.x versions

mrook:
We've sort of stopped with releasing 2.16.x versions. Is this a showstopper?

mrook:
It's only a warning, right?

Alexander1000:
yes, only warnings
not features

mrook:
Ok thanks. In that case I'm going to leave this one for now. We're working on a release candidate for 3.x, that has priority for us. Thanks for reporting!

Alexander1000:
how long to wait for release 3.x?

PR #1503, opened 2021-01-21
[php 8.0 compatability] deprecated ReflectionParameter::getType()

Alexander1000:
Patch for revision 2.x
This is break on php 8.0

@siad007 @mrook

### DIFF
--- a/classes/phing/IntrospectionHelper.php
+++ b/classes/phing/IntrospectionHelper.php
@@ -229,11 +229,7 @@ class IntrospectionHelper
                         );
                     }
 
-                    $classname = null;
-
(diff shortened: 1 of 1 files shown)


jawira:
I think this was already fixed with #1440

mrook:
That's right, but #1440 is against master (so 3.x).

mrook:
@Alexander1000 thanks. Based on this I think I need to roll a 2.16.4, I've also merged your other PR. Are there any more 2.x blocking issues that you've found?

Alexander1000:
Well, it seems so far everything that broke was found

PR #1545, opened 2021-03-10
Remove currentTimeMillis

jawira:
I remove `\Phing\Phing::currentTimeMillis` function since it can be easily replaced by 'microtime(true)'


siad007:
@jawira LGTM 👍🏼

PR #1560, opened 2021-03-29
[condition] Added phpversion condition

siad007:
Added phpversion condition:

```xml
<condition property="isTrue">
    <phpversion atleast="7.3"/>
</condition>
```


siad007:
Closed in favour of existing versioncompare condition.

reminder to myself - add a cookbook with this as an example 😄

PR #1583, opened 2021-04-21
Use constant for buildfile name

jawira:
I replaced the string `'build.xml'` by the appropriate constant, nothing more, nothing less.


jawira:
PhingVisualizer is blocking some checks, I will fix that 😬

jawira:
The CI error is due to package `mikey179/vfsstream` not being compatible with _PHP 8.1_.

More precisely the problem is this line:

https://github.com/bovigo/vfsStream/blob/7299174faac59e63a59c20cd775fbbb4f559d04e/src/vfsFile.php#L63

not respecting this rfc:

https://wiki.php.net/rfc/deprecate_null_to_scalar_internal_arg

jawira:
I reported the deprecation warning https://github.com/bovigo/vfsStream/issues/252

PR #1611, opened 2021-06-12
Update SizeSelector schema

jawira:
I removed `unit` property from `SizeSelector` schema to reflect changes made in https://github.com/phingofficial/phing/pull/1482

Also, `unit` property was already removed from doc https://www.phing.info/guide/chunkhtml/Size.html


siad007:
@jawira thx

PR #1627, opened 2021-08-19
Call vsprintf correctly in ChmodTask

bzikarsky:
`vsprint` expects its second argument to be an array. Old PHP versions silently also accepted a single non-array parameter that was "upgraded" to [$param].
This becomes an issue on PHP8, because it type-checks the parameter and issues an error accordingly.

This issue is already fixed in `main` and only exists in `oldstable`.

### DIFF
--- a/classes/phing/tasks/system/ChmodTask.php
+++ b/classes/phing/tasks/system/ChmodTask.php
@@ -188,8 +188,8 @@ class ChmodTask extends Task
         }
 
         if (!$this->verbose) {
-            $this->log('Total files changed to ' . vsprintf('%o', $mode) . ': ' . $total_files);
-            $this->log('Total directories changed to ' . vsprintf('%o', $mode) . ': ' . $total_dirs);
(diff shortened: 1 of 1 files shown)


siad007:
@bzikarsky thanks for your contribution but we're not planning on doing any more fixes to 2.x, sorry!

bzikarsky:
This leaves us in quite a hard place in terms of PHP8 compatibility. phing3 is not stable yet and phing2 does not work correctly with PHP8 for us because to the bug fixed in this MR.

mrook:
@siad007 given the number of PHP8 compat issues, we may need to consider rolling a 2.17 or something after all.

siad007:
> 
> 
> @siad007 given the number of PHP8 compat issues, we may need to consider rolling a 2.17 or something after all.

I am totally with you - would you mind to bundle one and would this compat issue be the only one? Feel free to chat me :)

mrook:
2.17.0 was just released.

PR #1636, opened 2021-09-01
Spawn open task

jawira:
Hello, I modified OpenTask a little bit:

- Everything works as expected when I open an url. But when I open a file, for example an image, the program blocks the terminal until the program is closed. That's why I added "spawn=true" option.
- I added "passthru=false" to hide any message in console.


jawira:
@mrook my fault, I will check that tonight.

PR #1643, opened 2021-09-11
Add ExecTask suggestions

jawira:
Display a suggestion when the executable from ExecTask can be replace by a Task.


jawira:
I think code is good to go, but I'm waiting your feedback. 
I will add some tests in coming days.

Issue #1641, opened 2021-09-10
3.0 pre-release checks


mrook:
This issue will contain a list of checks that need to be completed before we can mark 3.0 as stable and release it.

- [x] make it PHP 8.1 compatible including dependencies and tests (https://github.com/phingofficial/phing/issues/1642) 
- [x] finalize UPGRADING.md (https://github.com/phingofficial/phing/issues/1640)
- [x] verify the .phar contains all dependencies and works
- [x] deprecate `command` in `ExecTask`
- [x] document use of `phingVersion`
- [x] go through deprecations
- [x] versioned and stable phing/* (tasks etc.) dependencies rather than dev-master

mrook:
Refs #1662

siad007:
Bullet point `document use of phingVersion` is available through the docs.
For reference see
* https://www.phing.info/guide/hlhtml/#d5e938
* https://www.phing.info/guide/hlhtml/#Phingversion

siad007:
No deprecation warnings any longer (as far as I see) - will mark that point as done

Magiczne:
@siad007 are there any news about stable version 3? Currently version RC4 is conflicting with `dev-main` and `dev-master` branches in other packages from the ecosystem especially with composer-configurator.

siad007:
Right at the moment completely busy - but we have a look here (possibly at
next weekend) and will finish last issues regarding v3

Michał Kleszczyński ***@***.***> schrieb am Mo. 8. Aug. 2022
um 15:15:

PR #1647, opened 2021-10-30
Fixed PHP 8.1 issues

siad007:
Related to #1641


mrook:
@siad007 is this one necessary? I thought ~8.0 includes 8.1 as well (otherwise the 8.1 builds would fail?)

siad007:
@mrook unfortunately the builds for PHP 8.1 are failing with:
```
PHP Fatal error:  During inheritance of Iterator: Uncaught Return type of VersionControl_Git_Object_Tree::current() should either be compatible with Iterator::current(): mixed, or the #[\ReturnTypeWillChange] attribute should be used to temporarily suppress the notice
```
For reference see: https://github.com/phingofficial/phing/runs/4056016051?check_suite_focus=true

siad007:
Sorry - issues are gone - so closing it!

PR #1653, opened 2021-12-16
Type hasfreespace condition

jawira:
I added some typing and tests for HasFreeSpace condition


mrook:
@jawira can you update/finalize this one?

jawira:
@mrook Done. The idea behind this is MR is to throw an exception if path is invalid.

PR #1664, opened 2022-01-24
fix error: implicit conversion from float to int loses precision

Alexander1000:
implicit conversion from float to int loses precision in case when build work more than 1 minute
@mrook

### DIFF
--- a/classes/phing/listener/DefaultLogger.php
+++ b/classes/phing/listener/DefaultLogger.php
@@ -312,7 +312,7 @@ class DefaultLogger implements StreamRequiredBuildLogger
                 $minutes,
                 ($minutes === 1 ? " " : "s "),
                 $seconds - floor($seconds / 60) * 60,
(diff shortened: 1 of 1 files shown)


siad007:
@Alexander1000 thanks for the contribution, but AFAIK we do not make updates on the old-stable branch any longer. Is the issue still a problem in the current main branch?

Alexander1000:
But your main branch is still a release candidate. And the full version 3.0.0 has not yet been released. Because of this, we have to use version 2.x and this blocks the transition to a new version of php. As soon as you release version 3.0, we will immediately switch to it.

mrook:
@Alexander1000 is this a blocker for builds, or just a timing issue?

Alexander1000:
This breaks the build with an error in our ci/cd system.

mrook:
Ok. I'll push another stable release then.

PR #1702, opened 2023-04-04
Fix type of $outputDirectory parameter

isfedorov:
In SmartyTask::setOutputDirectory


mrook:
`\PhingFile` no longer exists, it's `Phing\Io\File`.

PR #1809, opened 2024-03-11
Prevent a PHP 8.3 deprecation notice

mpdude:
This prevents PHP 8.3 from complaining about a runtime-created property:

`PHP Deprecated:  Creation of dynamic property Phing::$searchForThis is deprecated`

### DIFF
--- a/classes/phing/Phing.php
+++ b/classes/phing/Phing.php
@@ -161,6 +161,9 @@ class Phing
     /** Whether or not output to the log is to be unadorned. */
     private $emacsMode = false;
 
+    /** @var ?string */
+    private $searchForThis = null;
+
     /**
      * Entry point allowing for more options from other front ends.
      *


mpdude:
Not sure about the release policy – this is against the `oldstable` branch, i. e. would probably go into a 2.17.x release? 

3.0 is still in the RC phase.

mrook:
Hi @mpdude I don't intend to do any oldstable releases. We're planning to a full 3.0 release somewhere in the next two months, depending on time & availability.

lpd-au:
This actually emits a deprecation notice since PHP 8.2, not 8.3. That means we're only a couple of quarters away from this project having no stable release that works on a PHP supported release. Given the patch is ready and trivial, I hope you might reconsider @mrook!

mrook:
Hi we're not going to do any more 2.x releases, the first 3.0 stable will be out soon.

PR #1816, opened 2024-04-23
(tasks/ext/git): Trim command output in setProperty

lea-BMS:
Fix missing trimming of EOL from command output in `setProperty`.


lea-BMS:
> Is this a duplicate of #1815 ?
Possibly, I was not sur into which branch to PR the changes. Thank you for review.

mrook:
The right branch is main.

PR #1820, opened 2024-04-27
Fix(Task/Ext/Archive/ZipTask): Preseves directories' permissions in zip

### LINKED ISSUE #1817: ZipTask stores directories with 777 permission
**Describe the bug**
When creating a zip archive with the Zip task, directory permission is 777 inside the archive.

**Steps To Reproduce**

```xml
<zip includeemptydirs="true" destfile="dist/${package}" basedir="./build">
    <fileset dir="build">
        <include name="**/**"/>
    </fileset>
</zip>
```

**Expected behavior**
Store original permissions bits or no extended attributes at all like it can be done with the zip command `-X` option

**Screenshots / terminal output**
If applicable, add screenshots or terminal output to help explain your problem.

list of extracted archive made with the ZipTask:
```sh
drwxrwxrwx 3 lea lea  4096 Apr 23 15:22 controllers
-rw-r--r-- 1 lea lea   302 Apr 18 16:00 index.php
-rw-r--r-- 1 lea lea   619 Apr 23 15:21 logo.gif
-rw-r--r-- 1 lea lea  1484 Apr 23 15:21 logo.png
drwxrwxrwx 2 lea lea  4096 Apr 23 15:22 sql
drwxrwxrwx 5 lea lea  4096 Apr 23 15:22 src
drwxrwxrwx 2 lea lea  4096 Apr 23 15:22 translations
drwxrwxrwx 5 lea lea  4096 Apr 23 15:22 views
```

This over permissive dangerous permissions on directory is not wanted at all.

**Additional context**

creating the very same zip archive with the `zip` command creates a zip archive that extracts perfectly well as expected:

```sh
drwxr-xr-x 3 lea lea  4096 Apr 18 16:00 controllers
-rw-r--r-- 1 lea lea   302 Apr 18 16:00 index.php
-rw-r--r-- 1 lea lea   619 Apr 23 15:21 logo.gif
-rw-r--r-- 1 lea lea  1484 Apr 23 15:21 logo.png
drwxr-xr-x 2 lea lea  4096 Apr 23 16:24 sql
drwxr-xr-x 5 lea lea  4096 Apr 18 16:00 src
drwxr-xr-x 2 lea lea  4096 Apr 23 16:24 translations
drwxr-xr-x 5 lea lea  4096 Apr 18 16:00 views
```

**Workaround with the `zip` command:**

```xml
<exec executable="zip" description="Create zip archive" dir="build">
    <arg value="-r9FSX" />
    <arg value="${project.basedir}/dist/${package}" />
    <arg value="${baseName}" />
</exec>
```

leagris:
Fix #1817 Stores directories's original permissions as `ZipArchive::OPSYS_UNIX` `ExternalAttribute`.

### DIFF
--- a/src/Phing/Task/Ext/Archive/ZipTask.php
+++ b/src/Phing/Task/Ext/Archive/ZipTask.php
@@ -306,6 +306,9 @@ class ZipTask extends MatchingTask
                 if ($f->isDirectory()) {
                     if ($pathInZip != '.') {
                         $zip->addEmptyDir($pathInZip);
+                        $dirAttrName = $pathInZip . '/';
+                        $dirAttrs = (int) fileperms($f->getPath()) << 16;
+                        $zip->setExternalAttributesName($dirAttrName, \ZipArchive::OPSYS_UNIX, $dirAttrs);
                     }
                 } else {
                     $zip->addFile($f->getAbsolutePath(), $pathInZip);


mrook:
Hi,

I'd like to make this behavior optional (i.e. controlled by an attribute). The attribute should default to false because we're changing behavior. The attribute should also be documented in the guide and the RNG grammar. When that's all done, it'll be added to the 3.1 release line (not 3.0).

leagris:
> Hi,
> 
> Thanks for your contribution! I'd like to make this behavior optional (i.e. controlled by an attribute). The attribute should default to false because we're changing behavior. The attribute should also be documented in the guide and the RNG grammar. When that's all done, it'll be added to the 3.1 release line (not 3.0).

I am ok with that, though, something annoys me with letting the current working as default.
With current behaviour, files permissions are preserved because it is handled natively by the `ZipArchive`'s `addFile` method. But since directories are not copied but created, ZipArchive does not know the original directory permissions.

Having files inheriting permissions but not directories ; it is more a bug than a feature. When dealing with such zip archives in a unix/Linux system, unzipped with the `unzip` command, those 777 read/write/execute permissions for everyone in directories is a serious problem. Letting this as the default feels quite odd.

I'd propose some alternative instead with a `savFileAttributes`  switch default `false` as you propose, to save the extra attributes (os-dependant or unix/POSIX only?)

The `ZipTask`'s `addFilesetsToArchive` method would not to be updated to implement those two attributes with:
- `savFileAttributes="false"` (**default**): Add files and directories to the zip without preserving their permission and timestamp. (Means not using `ZipArchive`'s `addFile` but some other method like [addFromString](https://www.php.net/manual/en/ziparchive.addfromstring.php) which does not have a source file attributes.
- `saveFileAttributes="true"`: The [addFile](https://www.php.net/manual/en/ziparchive.addfile.php) methods would do that natively, but it is not handled for directories. The handling of directories' file attributes need to be implemented outside the `ZipArchive`'s expansion which does not support it natively. It has OS-specific handling that can add complexity especially for OS/2 or VMs (does not look like a priority to deal with those). If we stick to Unix/Linux/BSD POSIX environment, permission bits and timestamps are the things to preserve in fileAttributes.

See: the [`zip` command's man-page](https://linux.die.net/man/1/zip) for the `-X`, `--no-extra` option:

> **`-X`**
> **`--no-extra`**
>     Do not save extra file attributes (Extended Attributes on OS/2, uid/gid and file times on Unix). The zip format uses extra fields to include additional information for each entry. Some extra fields are specific to particular systems while others are applicable to all systems. Normally when zip reads entries from an existing archive, it reads the extra fields it knows, strips the rest, and adds the extra fields applicable to that system. With -X, zip strips all old fields and only includes the Unicode and Zip64 extra fields (currently these two extra fields cannot be disabled).
> 
> Negating this option, `-X-`, includes all the default extra fields, but also copies over any unrecognized extra fields.

leagris:
@mrook Implemented a `savefileattributes` flag.

leagris:
Even Copilot says my code is smart ;)

> The `addFile` method in the `ZipArchive` class is typically used to add a file to the zip archive. However, in your script, it’s being used in a clever way to add a directory.
> 
> Here’s how it works:
> 
> When you call `addFile` with the `mtimeDummy` file and the `entryName` (which is the directory path), it adds an entry to the zip archive with the name of the directory. But because the `mtimeDummy` file is empty, this entry is effectively a directory, not a file.
> 
> The key here is the trailing slash in the `entryName`. In a zip archive, an entry with a trailing slash is treated as a directory.
> 
> So, even though `addFile` is normally used to add files, in this case, it’s being used to add a directory. This is possible because of the way entries are handled in zip archives and the use of the `mtimeDummy` file. This workaround allows you to preserve the `mtime` of the directory in the zip archive, despite the limitations of PHP 7.4’s `ZipArchive` class. It’s a smart solution! 👍

mrook:
I'll take a look at this soon. Sorry for not having had the time earlier.

siad007:
@leagris would you mind to add a test case please?
You can add one to this [BuildTest](https://github.com/phingofficial/phing/blob/main/tests/Phing/Test/Task/Ext/Archive/ZipUnzipTaskTest.php) and to the correspondent [TestBuild XML](https://github.com/phingofficial/phing/blob/main/tests/etc/tasks/ext/ZipUnzipTaskTest.xml) file.

PR #1829, opened 2024-05-02
Allow Symfony 7

holtkamp:


fredden:
Could we get a release with this change, please? I can see this has been added to a `3.1.0` milestone and that there are five other things in that milestone. Do we need to get all of those solved before a release will be considered? If so, are you looking for help with those?

PR #1838, opened 2024-06-03
add support for sebastian/version 5

TomasLudvik:
We are updating our dependencies in Shopsys Platform and PHPUnit 11 now requires sebastian/version in version 5, so this is a needed change so we can make this update, thank you.


siad007 (approved):
LGTM

mrook:
Duplicate of #1831 .

PR #1855, opened 2024-08-05
Update SonarTask.php

### LINKED ISSUE #1854: SonarTask: Tool banner no longer matches regex
Hey there,

Looks like 
https://github.com/phingofficial/phing/blob/f74ea08ee709c01e72ed3f816b8544a37674a653/src/Phing/Task/Ext/Analyzer/Sonar/SonarTask.php#L278

is out of date.

Current SonarScanner output (snippet)
```
01:13:03.756 INFO  SonarScanner CLI 6.1.0.4477
```
Perhaps it's good to add a flag to only made this a warning?

feamsr00:
quick fix to update banner. would fix #1854

### DIFF
--- a/src/Phing/Task/Ext/Analyzer/Sonar/SonarTask.php
+++ b/src/Phing/Task/Ext/Analyzer/Sonar/SonarTask.php
@@ -275,7 +275,7 @@ class SonarTask extends Task
 
         $isOk = false;
         foreach ($output as $line) {
-            if (preg_match('/Sonar(Qube )?Scanner \d+\\.\d+/', $line) === 1) {
+            if (preg_match('/Sonar(Qube )?Scanner CLI \d+\\.\d+/', $line) === 1) {
                 $isOk = true;
                 break;
             }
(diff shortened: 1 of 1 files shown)


src/Phing/Task/Ext/Analyzer/Sonar/SonarTask.php
```diff
@@ -275,7 +275,7 @@ private function validateExecutable()
 
         $isOk = false;
         foreach ($output as $line) {
-            if (preg_match('/Sonar(Qube )?Scanner \d+\\.\d+/', $line) === 1) {
+            if (preg_match('/Sonar(Qube )?Scanner CLI \d+\\.\d+/', $line) === 1) {
```
fredden, line 278: Do we need to support both old & new versions, or is supporting only the latest / new version of this tool correct?
```suggestion
            if (preg_match('/Sonar(Qube )?Scanner( CLI)? \d+\\.\d+/', $line) === 1) {
```

mrook: I have no idea when the tool banner was changed, and whether the version with the previous banner is still widely used.  The safe option would be to support old & new versions.

PR #1864, opened 2024-09-04
Allow new version of sebastian/version for phpunit 11 compatibility

tkegan:
This PR recreates PR #1831 which I mistakenly closed thinking it to be obsolete. If accepted, this PR updates the allowed versions of the dependency `sebastian/version` allowing `phing` to be used in projects which wish to use phpunit version 11.x for testing.

Note because `phing `uses `phpunit` version 9 for unit testing which depends on `sebastian/version` version 3.x running unit tests alone is not a sufficient test of this dependency change. Instead a project needs to be created using `phing` and `phpunit` (at version 11) and the `build.xml` needs to include a target which invokes `Phing::getPhingVersion`. I've tried this in a closed source project but can come up with a stripped down example if desired.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -65,7 +65,7 @@
         "ext-libxml": "*",
         "ext-simplexml": "*",
         "ext-xml": "*",
-        "sebastian/version": "^3.0|^4.0",
+        "sebastian/version": "^3.0|^4.0|^5.0",
         "symfony/console": "^5.3.10|^6.0|^7.0",
         "symfony/yaml": "^5.0|^6.0|^7.0"
     },


andresiri:
Running into the same problem as well, any estimation for when this will be merged @mrook 

In advance, i appreciate all the effort (:

siad007:
@tkegan Thanks for your contribution. Would you mind to include the generated `composer.lock` file as well?

tkegan:
> @tkegan Thanks for your contribution. Would you mind to include the generated `composer.lock` file as well?

Sorry I've been on call and it was kept hopping.

I'm not sure I understand. There are no changes to the `composer.lock` file required by this change. If I run `composer update phpunit` in the project for example, the checksum on line 7 of `composer.lock` ins changed but that's it. If I instead use `composer update` various dependencies are updated but I have not vetted those changes.

It may be helpful to see phpunit 11 coexisting with phing in a project; so I threw together a stripped down sample project you can try here https://github.com/tkegan/phing-sample.

Hopefully this helps. If not or there is something more I can do please let me know.

siad007 (approved):

bbouchard-seedbox:
@siad007 do you think it would be possible to create a version 3.0.1 ? We are locked on 3.0.0 which doesn't have the fix so it's impossible to upgrade to phpunit 11 still

PR #1886, opened 2024-12-04
Fixed failing dependabot workflow runs

siad007:
See for reference: https://github.com/phingofficial/phing/actions/runs/12104908943/job/33749537007

Added condition to ignore the dependabot actor for the docker build image step.


siad007:
@mrook after merge we could think about auto-merging the dependabot PRs then 😄 WDYT?

Issue #1889, opened 2024-12-30
Error on using switch statement "Value is required for case" 


Celebrate-Reinhard:
**Describe the bug**
After updating to phing/phing from 2.17.4 to 3.0.1, the switch statement throws an error.

**Steps To Reproduce**
```
<switch value="${WorkerNumber}">
    <case value="0">
        <echo message="The value of property WorkerNumber is 0" />
    </case>
    <case value="1">
        <echo message="The value of property WorkerNumber is 1" />
    </case>
    <default>
        <echo message="The value of property WorkerNumber is not sensible" />
    </default>
</switch>
```

There are no other changes to xml file. I am calling it as before like this:

`vendor/bin/phing -f /foo/build-prod.xml run-foo -DWorkerNumber=1`

**Expected behavior**
Echo statement of case 1 is visible.

**Screenshots / terminal output**
`build-prod.xml:12:40 Value is required for case`

**Additional context**
update of phing/phing from 2.17.4 to 3.0.1

PR #1899, opened 2025-01-14
[SwitchTask] Added test case for numeric case values

siad007:
Related to #1889


mrook:
@siad007 hey, noticed this PR is in draft - what's remaining?

PR #1933, opened 2025-07-09
updated symfony/filesystem dependency

jmalinens:


mrook:
This version pin is there to prevent using non-PHP8.4-compatible versions in the testing setup. It can ofcourse be amended to `^5.4.45|^6.4`.

PR #1940, opened 2025-08-28
 phing-grammar.rng: Fix optional attributes in different tasks

ramalama:
Previous definition did not capture how echoxml is working and documented:
Wrong was:
- If file or append attribute is supplied, the other one would be enforced
- If file or append is supplied, children are forbidden. Due to choice element attribute and children are mutually exclusive. 

Imho correct is: 
- file is optional
- append is optional - though makes no sense if no file attribute is supplied 
- any children can exist - no matter the variation of attributes

### DIFF
--- a/etc/phing-grammar.rng
+++ b/etc/phing-grammar.rng
@@ -1724,19 +1724,19 @@
 
     <define name="echoxml">
         <element name="echoxml">
-            <choice>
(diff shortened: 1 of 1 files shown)


ramalama:
Split into #1941 and #1942

Issue #1672, opened 2022-03-17
Disable output of some properties (passwords)


amigian74:
**Is your feature request related to a problem? Please describe.**
Right now all properties are logged with VERBOSE level on. This is critical in some situations, especially if passwords are involved (MySql for instance)

**Describe the solution you'd like**
It wolud be nice to add an attribute (hideoutput=true/false or disguiseoutput=true/false) to solve this problem.

mrook:
Hi, not sure I understand what you need. Do you want to change the log level of properties? Or mask some properties in all cases? For which tasks? In which situations? And why do you need to mask it?

amigian74:
Hi. The last one. With VERBOSE all properties used in a task are shown in the log. This also shows password properties (for instance for the pdosqlexec task). This is not so good - it would be better to mark such properties always hidden or disguise the output as usual for passwords.

mrook:
This is not trivial. Tasks are not in control of the dumping of properties to console under `VERBOSE`, that's a core Phing capability. We'd need to either add whether a property is sensitive or not, or use regexp masking rules to replace passwords with other characters when under `VERBOSE`.

Having a "do not dump properties" option while running phing in debug mode is also an option, but feels like a shortcut to fix a symptom.

I'm not entirely sure we need this yet, interested to see what other users think.

mrook:
@siad007 WDYT?

MiRacLe-RPZ:
For my custom-tasks i do something like:

```
    public function log($msg, $level = Project::MSG_INFO, ?Exception $t = NULL) {
        $msg = preg_replace('~--password="?([^"]+)"?\s~', '--password="******" ', $msg);
        $this->project->logObject($this, $msg, $level, $t);
    }
```

No idea how to do this globally, but per task we can hide sensitive data by this way.

siad007:
We could simply introduce a logger, which could hide by configuration either

* all properties or
* properties by regex or
* some global definable things like user/password from url etc.

jawira:
I was reading this RFC [Redacting parameters in back traces](https://wiki.php.net/rfc/redact_parameters_in_back_traces) and I remembered this issue. Maybe it can be useful, at least as inspiration?

mrook:
Coming back to this, we can improve the `DisguiseLogger` by adding additional regular expressions to mask various secrets. Anyone willing to pick that up?

jawira:
Hello @mrook I can help improving the `DisguiseLogger` :) please give me more details.

PR #1977, opened 2025-12-28
Mask Github tokens in DisguiseLogger

jawira:
Mask well known patterns used by GitHub tokens.

See #1672


jawira:
Hello @mrook, this PR do not has test because I cannot find a way to execute the tests with DisguiseLogger. One alternative is to extract masking strategies into classes - one class per strategy - and then test the public masking method as unit tests. Let me know what do you think.

